PR #1, opened 2024-10-08
Fix crash issue from a different mod

Jevonnissocoolman:
Make these tarots check if G.jokers exists first before checking for their specific jokers
This prevents a crash from a different mod on launch
I tested this and the in_pool should still be functionally the same


InertSteak:
Thank you! That's a good sanity check to have in place.

PR #3, opened 2024-10-12
Added art to atlases and set up soul pos on legendaries

GayCoonie:
title


InertSteak:
Oops, that's what I get for making a hasty commit. Thanks guys!

PR #5, opened 2024-10-13
Update lovely.toml

lshtech:
better compat for rarity badge


InertSteak:
Thanks! I was meaning to get to that, this is a big help

PR #25, opened 2024-11-02
dynamic farfetchd text

jo-z:
Farftech'd's example text wasn't updating when it was leveled with energy (e.g. 3.5X per leek still saying 2 leeks would get you 6X). I added a loc_var and referenced it to fix this


InertSteak:
Yeah so unfortunately I just reworked farfetchd so this won't be needed. Sorry about that, but I really appreciate the contribution!

PR #45, opened 2024-11-10
Remove "Thunder Stone" tooltip on Pichu

SDM0:
Thunder Stone isn't mentioned for Pichu.


InertSteak:
Yup, that's an oversight. Thanks!

PR #59, opened 2024-11-20
Fix Venonat line jank with Oops! all 6s

Jevonnissocoolman:
stumbled across this while playing this mod and it was extremely jank so I fixed it

note: you will still need to account for any other jokers from this mod or other mods that mess with probabilities


InertSteak:
Thanks! Yeah its kind of a total pain with the probability stuff, I'll keep that in mind

PR #67, opened 2024-11-23
Gen 2 Sprites

Numbuh214:
I think this should be everything? Lemme know if there's something I missed.


InertSteak:
Awesome! Will do

PR #70, opened 2024-11-24
Added Numel And Camerup in the atlas

YamperAF:
For very future use, i just made it because the lava background of pokemon mistery dungeon looks cool and my mind says, do camerup and numel, and there is are this two volcano guys


InertSteak:
Cool! I really like them, makes me start thinking about what their effect would be

PR #93, opened 2024-12-25
@inertsteak

Jevonnissocoolman:
scuffed af method of applying shortcut effect to rapidash

find_joker >>>> SMODS.find_card fr fr


InertSteak:
That's my kinda crazy lol

PR #96, opened 2025-01-03
Machop line fix when debuffed

Jevonnissocoolman:
Fixes (what I assume to be) incorrect behaviour when debuffed

### DIFF
--- a/pokemon/pokejokers_3.lua
+++ b/pokemon/pokejokers_3.lua
@@ -243,18 +243,18 @@ local machop={
     return level_evo(self, card, context, "j_poke_machoke")
   end,
   add_to_deck = function(self, card, from_debuff)
+    G.GAME.round_resets.hands = G.GAME.round_resets.hands + card.ability.extra.hands
+    G.GAME.round_resets.discards = G.GAME.round_resets.discards - card.ability.extra.discards
     if not from_debuff then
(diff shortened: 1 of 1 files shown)


InertSteak:
The reason it doesn't do it on debuff is to fix an issue where you would get a bunch of hands from a boss that buffed and debuffed Jokers. I don't think this is how we want it to behave

Jevonnissocoolman:
yeah I saw that and that's what I assumed and it still won't do that I think
The passive hands/discards per round don't get added/removed on debuff so if it gets debuffed it won't have an effect on future rounds which Is what I fixed

InertSteak:
I'm not sure I follow. What problem is this change fixing?

Jevonnissocoolman:
Uh the G.GAME.round_resets_hands stuff (which changes the hand/discard amount) right now doesn't get changed if the joker is debuffed which isn't supposed to happen and can potentially cause the joker to do nothing so I moved it out of the if statement
the stuff that gives the hands mid round I left in there since I assume that's the behaviour you wanted to stop

InertSteak:
Ah I see, alright I'll go ahead with the merge, sorry for the confusion!

InertSteak:
Just fyi, the same problem that I fixed before cropped up again with this change. That being Machamp granting additional hands. So I'm reverting the change until we can think of a better solution

PR #119, opened 2025-02-03
Atlas Update Part 2

Lemmanade0:
Every pokermon now uses the proper yellow art border, Litwick line significantly brightened background to make it look better in-game.


InertSteak:
Looks great, thanks!

PR #121, opened 2025-02-04
Lemma Atlas Update #3

Lemmanade0:
All the updates I already did but on the not-numbered atlases


InertSteak:
Doing Arceus's work here

PR #125, opened 2025-02-05
Gen 4 Complete Atlas

SonfiveTV:
Froslass was in the wrong place so updated the location in the code

Gen 4 in its entirety, I didn't included a numbered version because we are phasing those out, but I included a renamed copy of the original so people don't have missing textures for the time being


InertSteak:
Awesome! Thanks so much

PR #126, opened 2025-02-06
Automated Pinsir's check for matching cards

random-facades:
Pinsir now gives the mult, if there are matching cards in the played card and the hand


InertSteak:
Merging this now

PR #128, opened 2025-02-07
Gen4 Atlas Fix

SonfiveTV:
I had left some artifacts on some jokers, see Riolu as an example

Should all be fixed now.


InertSteak:
Sounds good, thanks!

PR #132, opened 2025-02-09
Egg

random-facades:
Hatches after 2-4 turns.
Costs $1 and sells for $1

Upon creation, a basic or baby pokemon is hidden inside, which has a 1% chance to be a legendary instead. The hatch time is based around the rarity, with common hatching in 2 rounds, uncommon/rare hatching in 3 rounds, and legendary in 4 rounds.

Additionally, pokemon with Egg Helper abilities will speed up the hatch time slightly.
Without any helpers: 2/3/3/4
With one helper: 2/2/3/3
With two helpers: 2/2/2/3

If the egg is shiny, the hatched pokemon will be shiny, and non-shiny eggs have a greater chance to be shiny.

Mystery Egg


random-facades:
Changing Mystery Egg based on feedback in the suggestions channel

- Increased duration of egg
- Adjusted helpers to only be Flame Body, Steam Engine, and Magma Armor
- Added shaking when it's ready to hatch (like evolution)
- Increased shiny chance
- Changed selection to match how Pokeball functions (equal chance regardless of rarity)

random-facades:
image

Got the changing text working
It should pull from localizations, so we don't have to have the same number of descriptions for each region.

InertSteak:
I'm interested to bring this in with a few changes. 
- Nix the chance on legendary as it doesn't really work that way in pokemon
- Instead have a higher chance for an edition or +1 energy or both. 
- Also make it increase in sell value every round like base balatro egg. I was thinking it would increase by 1$ up to a cap of $6 and it would hatch when you sold it after the 2-4 rounds.
Otherwise I like everything going on here

random-facades:
Currently, the egg has a chance for an edition (which will be copied to the hatched mon), and if the egg doesn't have an edition, the hatched mon has a chance to gain one as well.
I added a 10% chance to gain an energy (if the mon can gain energy)
I removed the legendary possibility, and now it'll only be Baby/Basic

To add the sell value/sell trigger, it'll need some reworking as a sell trigger will need the description to explain how to use it, and the random hatch time will be difficult to explain with a sell trigger.

Let me know how you'd like to proceed

PR #134, opened 2025-02-11
Oricorio & Pokermon Forms

random-facades:
Oricorio - Rare ($10)
X4 Mult if played hand contains a Flush with <SUIT>
Changes form if played hand contains a different Flush

It won't score when changing forms, so it's difficult to score twice without major deck manipulation.

image

This joker is only one joker with four sprites based on the internal values. The sprites are changed upon load, and the `family` table was modified to support displaying different forms. Basically, each pokermon with forms will have an internal form value and an `update_form` function to support changing of forms.


andyape (approved):

random-facades:
Closing due to main branch having Forms logic.
Will readd, when oricorio design is approved.

PR #135, opened 2025-02-12
Celebi

random-facades:
Celebi - Legendary ($20)

Decreases Ante by 1 for every 4 Blinds skipped
Could be "infinite" with two.

Have two descriptions that could be used. Not sure which is better:

image image


random-facades:
Celebi added through alternate method

PR #136, opened 2025-02-12
Gen V Starters

random-facades:
All Gen V starters (without sprites)
Focused around discards.
Has the possibilities to be too synergistic, but should be limited by the fact that only the Basic will be in shops, and more pokemon should prevent getting too many.

oshawott_line
tepig_line
snivy_line


InertSteak:
Closing old PRS

PR #140, opened 2025-02-15
Perma Mult w/ Slugma & Magcargo

random-facades:
Added functionality to support permanent Mult upgrades:

Slugma: ($3 Common)
> At start of round, add a Stone card to hand for each adjacent Water Joker
> Evolves after creating 5 Stone cards

Magcargo: ($7 Rare)
> Stone cards game +1 Mult permanently when scored for each Fire Joker you have


random-facades:
Gonna remake this with updated smods

PR #142, opened 2025-02-16
Misc Fixes

random-facades:
Torchic issue
> Always says "+1 Mult", even with other fire/fighting types

Magneton issue
> Will look at left type twice

Lairon/Aggron issue
> Starts at X2/X4 mult when purchased/found


InertSteak:
Looks good, thanks!

PR #143, opened 2025-02-16
Professor Jimbo w/Easter Egg

random-facades:
Jimbo (Uncommon $10)
> Gives a Starter and Pokedex when sold
> Starter and Pokedex sells for $2-$3 each, and selling Jimbo gives $5, selling everything gives $9 - $10

image


InertSteak:
Closing as it has been implemented a bit differently

PR #145, opened 2025-02-17
Gen 9 atlas

princessroxie:
I'm pretty sure I did everything right. I had to adjust the positions of tinkatink/ape/gimmighoul in the code.


InertSteak:
Awesome, thanks!

InertSteak:
Only problem was Gimmighoul(roaming) didn't have the correct pos, but I fixed that. Everything else looks good

PR #149, opened 2025-02-18
Force crash if lovely.toml doesn't load

random-facades:
~~Checks `SMODS.calculation_keys` for `'xchips'`
If `'xchips'` isn't present, crash and tell them to update SMODS~~

Checks for `poke_safari` colors in the LOC_COLOURS table.
If `poke_safari` doesn't exist, the lovely.toml didn't get loaded, so crash with message "POKERMON - INCORRECT FILE STRUCTURE"


InertSteak:
Looks good, merging

PR #153, opened 2025-02-20
Allow Forced Evolution

random-facades:
Evolution items can be used for only evolution, if the highlighted jokers use the evolution item.

If the item is highlighting cards properly (i.e. thunderstone with 1 card highlighted), then it will behave normally, even if a joker is highlighted.


InertSteak:
Thanks so much, I'll look into merging this tomorrow

PR #156, opened 2025-02-20
Gen 1-3 atlas fixing

princessroxie:
Changed Mythicals to have an M instead of an L, added shiny borders, fixed some mistakes on the gen 1 shiny atlas, fixed octillery not being on the shiny atlas for some reason? and also added some new castform art that doesnt obscure castform as much and looks better imo (thanks to sonfive for the castform)


InertSteak:
Awesome, thanks for all the hard work!

PR #157, opened 2025-02-20
Gen 4 shiny borders

princessroxie:
I also made Sunny Cherrim's background into the sunny forest one since I think it fits


InertSteak:
Nicely done

PR #158, opened 2025-02-20
Jirachi forms atlases

princessroxie:
Obviously the code for the card itself isn't done, but these should probably be in the files for whenever we do get around to that


InertSteak:
Good thinking. Is there a reason there's only the 2x version? Shouldn't there be a 1x version as well?

princessroxie:
I uploaded 1x and 2x versions? Did I do it wrong? It shows both on my end

InertSteak:
Sorry you're good. I goofed and clicked on the wrong thing lol

PR #162, opened 2025-02-21
Unified naming conventions for Pokejokers files.

Catzzadilla:
This should help make things more uniform. It's in the same order, this just gets rid of the WXY in front of them.


InertSteak:
Looks good, merging

PR #169, opened 2025-03-01
Evo Juice Load Fix

random-facades:
Evolution jitter doesn't get loaded when restarting a run
Removed existing jitter code within a few jokers and updated pokefunctions to better track evolution logic


InertSteak:
Looks good, merging

PR #170, opened 2025-03-01
Resolve Inconsistent add/remove to deck logic

random-facades:
Some pokermons would decrease value when debuffed, but not increase them properly, leading to a permanent loss of hands/discards/hand size

Matched all the functionality to base Balatro (Troubadour, Juggler, Drunkard)


InertSteak:
Thanks, this is a big help!

PR #173, opened 2025-03-02
Calculate Function Nil fix

random-facades:
If `self.calculate` was nil, the game will crash on load as it tries to run on load.


InertSteak:
Got it, merging

PR #174, opened 2025-03-02
Legendary outlines fix

princessroxie:
Making it more consistent with the other gens, the non shiny legends borders for gen 2 weren't the right color


princessroxie:
i noticed more problems so i fixed those too, including shiny outlines and a cleffa/pichu text fix

InertSteak:
Awesome, thanks!

PR #175, opened 2025-03-02
More gen 2 atlas fixes

princessroxie:
haha you ever just notice more problems after you already fixed another one?

Specifically fixed togepi and togetic name position, and a stray pixel that shouldve been transparent but was not


InertSteak:
All the time lol

PR #177, opened 2025-03-03
Pokedex Change

random-facades:
Pokedex:
> +1 Mult for each unique Joker seen
> Pokemon from the same evolutionary line may appear

image

When a Joker is created mid scoring, it should not update the mult until after the scoring.
Pokedex observers jokers in the shops slots, joker slots, or pack slots (when a booster pack is opened)

### DIFF
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -2878,8 +2878,7 @@ return {
             j_poke_pokedex = {
                 name = 'Pokedex',
                 text = {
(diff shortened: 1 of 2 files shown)


JTBringe:
I might be late to the party here, but I like the concept 😄
If I could suggest an addition it would be to perhaps add more mult or a small x mult for every joker you've "caught" (had in your lineup) this run.

InertSteak:
Closing as pokedex has been tweaked already

PR #183, opened 2025-03-04
Evolution Text Review

random-facades:
Updated all Evolution text to fit within these categories:

Level Evo: (No change)
> Evolves after X rounds (X decreases every round)

Scaling Evo (Non-scoring tracker):
> Evolves after <doing> X <things> (X decreases when <doing> <things>)

Scaling Evo (Basically anything with a currently):
> Evolves at <Current> / <Target> <Chips/Mult>

Item Evo:
> Evolves with a <Item>

Exceptions:
> There's always exceptions
> Any effect where it can go up and down (like seals in deck)


InertSteak:
Doing arceus' work

PR #197, opened 2025-03-08
Zorua Line

random-facades:
Zorua (Uncommon $10):
> Copies ability of rightmost Joker
> Illusion breaks after first hand
> Evolves after 5 rounds
If you don't win in one hand, Zorua faints (like babies and volatile) and doesn't level up.

Zoroark (Safari $12):
> Copies ability of rightmost Joker

Both Zorua and Zoroark will be hidden inside the shops and disguised as other pokemon:
image
image

After purchase, it'll show it's proper tooltip, but will be disguised as the target if it's blueprint compatible:
image
image


random-facades:
Fixed missing `rental_compat` functionality

PR #199, opened 2025-03-09
Natu Line

random-facades:
Natu (Common $4):
> Planet cards give an extra level
> Evolves after 4 rounds

Xatu (Uncommon $7):
> Planet cards give 2 extra levels
> Celestial Packs contain every Planet card

natu line

Also, found an issue where the pack size was getting overridden by SMODS. When pack is opened, the `booster.ability.extra` is updated and SMODS will use this value to size the pack UI and to determine the amount of card in the pack.


InertSteak:
Approved after a rarity change!

PR #200, opened 2025-03-09
Family Checks during card creation

random-facades:
When creating the pools of possible cards, any pokemon will go through a "family check".
The "family check" simply looks through the family lists for this pokemon, and then looks if this pokemon is present in the G.GAME.used_jokers table, which is where the game stores temporary jokers (i.e. shop jokers)


random-facades:
Ah, this includes the drakloak fix too, whoops

PR #201, opened 2025-03-09
Gen 2 & 3 Sprite Cleanup

SonfiveTV:
Some of the sprites had effects applied to them but this made them stick out from the rest and reduced visibility, so I made them more consistent with the rest of the Jokers


InertSteak:
Awesome, thanks for all the work!

PR #202, opened 2025-03-10
Gholdengo Values to Big with Talisman

random-facades:
Gholdengo is currently just basic `number` values, which are limited by lua's values of around e308.
With this change, Gholdengo's values will change to Talisman's Big value upon any calculation


random-facades:
Tested it with Mystic~Rain, and it appears to be working

InertSteak:
Thanks for the fix!

PR #204, opened 2025-03-11
Zorua fix

random-facades:
Was unable to determine what could be causing the name table to not be initialized.
I forced the creation of it for all hidden Zorua, which should stop the crashes.


InertSteak:
Sounds good, thanks!

PR #220, opened 2025-03-16
Zorua Legendary Fix

random-facades:
Zorua/Zoroark now copy the `floating_sprite` from the rightmost joker, if it has one


InertSteak:
Nice! The sprite based stuff always gives me trouble

PR #222, opened 2025-03-16
Unown as Playing Cards

random-facades:
image

Unown act similar to Stone cards - Always Score and are worth flat chips
However, they can then be enhanced/editioned/sealed like normal cards

One unown is set to be `in_pool` and can be found in standard card packs.
After adding it to deck, the others appear and can be found in standard card packs.

Six unown have special scoring effects:
- ! and ? give 20 mult and 100 chips respectively, if they're the last cards played (can play YES!! and both will count)
- I, V, X, and L give chips equal to their roman scoring values (1, 5, 10, and 50 respectively)

Other unown have special effects when spelling a word out.
All the words are present in the `unownrewards.lua` and can be added to fairly easily


random-facades:
I think this is ready to review!
There's... a lot of stuff in here

InertSteak:
Closing old pull requests

PR #230, opened 2025-03-18
JokerDisplay support for Gen 1

SonfiveTV:
All of Gen 1 is implemented, as well as the odd few from other gens

I will get round to doing the rest eventually, but for now I chill


SonfiveTV:
Oh btw Wigglytuff currently doesn't account for the played cards' total chips, because I couldn't figure out how to do that

InertSteak:
Dude, I can't thank you enough for this. Stellar work!

PR #232, opened 2025-03-19
Pawniard and Bisharp changes

princessroxie:
Changed their type to Steel, to match Kingambit, and also added a "Pawniard Army" and "Bisharp General" card.


InertSteak:
Alright sounds good!

PR #235, opened 2025-03-20
Articuno JokerDisplay removed

SonfiveTV:
I thought it only triggered on the first hand, it triggers every hand so jokerdisplay isnt needed for Articuno


InertSteak:
Gotcha, merging!

PR #246, opened 2025-03-25
Hazard Draw Fix

random-facades:
Hazard cards trigger individual draws upon load/draw
Will trigger a calculation after finishing drawing
Swapped jokers to checking only the drawn cards instead of the whole hand


InertSteak:
Nicely done

PR #248, opened 2025-03-25
A few more jokerdisplays

SonfiveTV:
Not that many here but I wanted to submit (I will try to do more per PR going forward so I'm not bombarding you with little updates)

Adds:
Hoothoot but is commented out for now because it doesnt count retriggers so not entirely accurate
Gimmighoul
Gholdengo
All the "Other" Jokers, except Rival is commented out because the forms changing between +mult and Xmult breaks it slightly

### DIFF
--- a/definitions1.lua
+++ b/definitions1.lua
@@ -1,0 +1,3063 @@
+local jd_def = JokerDisplay.Definitions
+
+
+jd_def["j_poke_bulbasaur"] = {
+    text = {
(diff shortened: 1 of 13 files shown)


SonfiveTV:
Skarmory and Roggenrola line would crash the game if JokerDisplay is enabled because of the new changes, I have fixed them now

InertSteak:
Good stuff! Merging

PR #249, opened 2025-03-26
display fixes

SonfiveTV:
Removed Tarountula line's displays, not needed now
Fixed Dugtrio & Wugtrio showing 0X instead of 1X
Fixed Litleo/Pyroar


InertSteak:
Nice, thanks!

PR #252, opened 2025-03-27
Duplicate file removal

SonfiveTV:
I noticed I had accidentally uploaded files to the root folder instead of into jokerdisplay so I've removed these duplicates


InertSteak:
Yeah I noticed that lol

PR #255, opened 2025-03-28
Add Shinx line

artemisSystem:
Based on idea by @catra1234 on discord, with slight modifications.

Every 1/8, 1/6 or 1/2 cards (for Shinx, Luxio and Luxray respectively) are drawn face down (like the wheel boss blind).
Shinx (common) gives $1 per scored face down card.
Luxio (uncommon) gives $1 and +2 Mult.
Luxray (safari) gives $1 and X1.1 Mult. Additionally, the hand type preview shows even if face down cards are selected. (Reference to Luxray's x-ray vision)

Both shinx and luxio evolve after 3 rounds.

Tested:
- Works with Oops all 6s
- Works with energy

Something i noticed: When drawing cards for booster packs in the shop, cards can still be drawn face down. Should maybe be fixed because the face down cards don't give any advantage there.

This is my first time modding balatro, please let me know if anything seems wrong or can be done in a better way.

Open to suggestions about tweaking the numbers of course, they might need to be tweaked a bit to fit in with the balance of the rest of the jokers.

### DIFF
--- a/functions/pokefunctions.lua
+++ b/functions/pokefunctions.lua
@@ -99,6 +99,7 @@ family = {
     {"mudkip", "marshtomp", "swampert"},
     {"shroomish", "breloom"},
     {"aron","lairon","aggron"},
+    {"shinx", "luxio", "luxray"},
     {"buizel", "floatzel"},
     {"gothita", "gothorita", "gothitelle"},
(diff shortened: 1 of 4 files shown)


InertSteak:
Closing old PRS

PR #258, opened 2025-03-29
Hoothoot line Stone Card fix & Mantine Visual Fix

random-facades:
Hoothoot will give chips equal to `nominal + bonus + perma_bonus` for all card types, including Stone Cards.
For Stone Cards, the `nominal` chips should be ignored.
This fixes that

ALSO:
Mantine previously has a double "square" appear when a gold card in hand is triggered.
This removes that extra square


InertSteak:
Awesome, thanks!

PR #259, opened 2025-03-29
Zorua Shader and pokedex fix

random-facades:
Added a shader for only Zorua/Zoroark while in the shop to help visibility
zorua example

Also, fixed the right click menu not showing Zorua/Zoroark.
Also, added a dark sticker to both so their types are visible even when stealthed


InertSteak:
Amazing work!

PR #262, opened 2025-03-31
Added Wish Pack sprite

Catzzadilla:
Added a Wish pack sprite for when a wish is made with Jirachi.

~~also made a change to the alcremie line sheet because I think a kitchen background looks better than brick~~


InertSteak:
Wicked

PR #263, opened 2025-03-31
Egg Joker Functionality

random-facades:
Allow setting egg target and prevent egg logic from overwriting the set value.
Can be used with Daycare Jokers to spawn specific eggs


InertSteak:
Good change

PR #264, opened 2025-03-31
JokerDisplay batch

SonfiveTV:
These are all that are left, but I have no idea how to do them so for now, I am done with JokerDisplay additions:

Wigglytuff - Doesn't calculate Total Chips for each card played
Hoothoot - Is commented out because doesn't count retriggers in scry area
Noctowl - Is commented out because doesn't count retriggers in scry area
Remoraid - Not sure how to implement retrigger count for these
Octillery - Not sure how to implement retrigger count for these
Metagross - Only displays +chip value, doesn't calculate Xmult using played cards total chips
Jirachi - No comment your honor

### DIFF
--- a/jokerdisplay/definitions2.lua
+++ b/jokerdisplay/definitions2.lua
@@ -329,6 +329,48 @@ jd_def["j_poke_togepi"] = {
 --	Flaaffy
 --	Ampharos
 --	Bellossom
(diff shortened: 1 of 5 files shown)


InertSteak:
Awesome to hear! Merging!

PR #267, opened 2025-03-31
Fixed a bunch of errors

princessroxie:
Spoink had a great ball, one of the backgrounds had a minor mistake on every atlas it appeared on, + a few other tiny things


InertSteak:
Looks good, merging

PR #272, opened 2025-04-01
JokerDisplay

SonfiveTV:
I somehow missed Buizel + Floatzel, and Shellder/Cloyster were counting the hand size, not the scoring hand size


InertSteak:
Merging!

PR #275, opened 2025-04-01
Zorua Shader applies when it is stealthed

random-facades:


random-facades:
Whoops, wrong logic.
Give me a moment to fix

PR #279, opened 2025-04-03
Smeargle "Canvas"

random-facades:
Smeargle now has a floating sprite that is the copied joker, if there is one.
It also supports Animations!
Examples below:
smeargle unown
image
image


random-facades:
Gonna close this to work on it some more

PR #286, opened 2025-04-05
Used jokers fix

random-facades:
There's an issue that's occurring when you "transform" a joker into another:
> It doesn't get removed from the `used_jokers` table, and is being treated as if it's "on-screen".

This causes evolution to not remove the previous pokermon from the `used_jokers` and will never spawn in the shop.

This should be fixed with these changes.


InertSteak:
Great work!

PR #292, opened 2025-04-07
Tinkaton shouldn't break debuffed steel cards

english5040:


english5040:
Tinkaton's code is a bit messy, it should probably destroy cards using smods' context.destroy_card

InertSteak:
Yeah, instead of merging, let me refactor it. I didn't know about that context while coding it

InertSteak:
Change of plans, the reason Tinkaton looks like this is to have steel cards mimic glass, which is a bit harder to do with context.destroying_card. I'll just merge your change

PR #294, opened 2025-04-08
Mareep Line

random-facades:
mareep_line

Mareep is a $4 Common, which may lead to rapid evolution.
To alleviate this, Mareep's sell value gets set to it's cost when purchased (without coupons), and is incompatible with rental. However, it still will evolve in 1 round if played correctly.

Additional code changes:
- `poke_drain` was refactored to be `poke_drain(drain_target, amount, gain_target)` and will return the amount drained
- `money_mod1` and `money_mod2` were added to the energy calculations


random-facades:
Forgot about Mega Ampharos, I'll get that spun up

InertSteak:
Closing this as we ended up adding a completely different version

PR #305, opened 2025-04-11
Custom Megas Compatibility Fix

LegendDuoKeeper:
I'm still pretty new to coding and github etiquette, but I think I'm doing this close enough to correctly? Anyways, I tested these changes on my own custom mega evo, and multiple native mega evo pokermon. Activating, deactivating and selling a mega stone, with any combination of custom and native megas, in various combinations of positions, all appear to be working as desired. (8 commits are listed for what should be 2 small changes due to me running into a very niche bug after opting for some last minute testing right before I initially went to submit this. Said bug has been fixed now.)

### DIFF
--- a/consumables/mart1.lua
+++ b/consumables/mart1.lua
@@ -576,6 +576,7 @@ local megastone = {
   end,
   use = function(self, card, area, copier)
     local target = nil
+    local forced_key = nil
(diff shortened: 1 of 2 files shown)


InertSteak:
I was literally about to tackle this when I see I already have a nice and neat PR. Thanks a bunch! Looks like you did everything right to me.

PR #311, opened 2025-04-19
Implement roxie's Drifloon and Drifblim

marcavis:
Notes:
-Vouchers need some investigation: Pokermon vouchers have the unredeem function to undo their effects,
but vanilla vouchers don't. What to do?

-Must check the TODO: in drifloon; when a new joker is bought, Drifloon/blim is bumped out of the rightmost position, so checking for volatile_active didn't work.
Isn't the solution of checking whether the new joker is at G.jokers.config.card_count + 1 position a bit wonky?


marcavis:
Ah yes, I didn't do anything with booster packs yet. They don't call the buying_card context, I am not too sure whether they should count, since you don't know what you'll find inside.

InertSteak:
Closing old PRS

PR #315, opened 2025-04-23
Fix: Evolving shinys no longer adds an extra booster pack

DecoXFE:
Previously, evolving a shiny Pokémon would incorrectly increase the booster pack limit. This fix prevents that by checking if the card is already shiny before applying the effect.


InertSteak:
This had already been fixed, so I'm closing this pr.

PR #318, opened 2025-05-01
Update zh_CN localization

ayssia:
Some wording might not be accurate. Feel free to make further corrections.


InertSteak:
Awesome, merging!

PR #324, opened 2025-05-10
fix booster pack/ĩtem in jokers area interaction

marcavis:
With this change, whenever a booster pack is open and an item is used, items in the Jokers area are also exempted from reducing the number of booster pack choices left.


InertSteak:
Simple but effective. Nice!

PR #329, opened 2025-05-15
JP translation Swinub and Yanma line update

khanryss:
Also fixed some missed untranslated text, and the updated Dunsparce.


InertSteak:
Nice, thanks!

PR #331, opened 2025-05-16
Add Joker Display definitions for Snubbull, Azurill, and Gligar evo lines. Fix Dragonite's Joker Display bug

Definite-Internet-User:
Dragonite's jokerdisplay definition had a bug where it would apply its five retriggers to cards held in hand and that caused jokers like Golurk to display inaccurate joker display information.


InertSteak:
Awesome, great stuff!

PR #345, opened 2025-05-31
Make Energy cards properly display max energy

gold-me:
pretty small fix


InertSteak:
It ended up being easier to manually do it with the conflicts, but thanks for the solution!

PR #353, opened 2025-06-02
French localization for everything but jokers

Maelmc:
Some jokers are partially translated becasue of "Replace All"


InertSteak:
Nice!

PR #357, opened 2025-06-04
Joker Display Pichu bug fix

Definite-Internet-User:
Pichu joker display's money wasn't updating when energized. Also I decided to put the xmult in the same line with +$ since being able to collapse the xmult was a bit weird and to bring it in line with other jokers that simultaneously give xmult and money.


InertSteak:
Makes sense, good work!

PR #360, opened 2025-06-05
Translating all pokemon names from English to French

Toadsword:
Translated all pokemon names in French


InertSteak:
Nicely done!

PR #374, opened 2025-06-13
Fix stantler trigger

six6game:


InertSteak:
I think this changes the functionality we're going for, so I'm closing this

PR #382, opened 2025-06-19
Replacing function calls with SMODS.has_enhancement() where applicable for cross-mod compatibility

Eternalnacho:
As title suggests - also reformatted the comment lines for the first five pokejokers files to match pokejokers_06 and beyond


InertSteak:
Wow, you're a superstar. Thanks so much!

PR #384, opened 2025-06-20
Amped Sleeve bug fix

SonfiveTV:
Amped Sleeve would apply twice, think it was because it had the same key as the deck? Either way, I changed the name and key, and the bug disappeared


InertSteak:
Oh yeah whoops that was my bad, thanks!

PR #385, opened 2025-06-20
Fixing a Chansey Bug

Eternalnacho:
Pokejokers_04, called Smods_has_enhancement on context.other_card instead of v and it indexed wrong and crashed whoops


InertSteak:
No worries, happens to the best of us

PR #386, opened 2025-06-22
Phanpy blueprint fix

SonfiveTV:
If blueprint'd, Phanpy's Xmult was getting reset regardless of whether you played 5 cards or not


InertSteak:
Thanks for pointing this out! I had fixed Donphan but not Phanpy. I went ahead and made the change on my own end however, your merge has a syntax error in it, missing the and between logical statements.

PR #388, opened 2025-06-25
Hazard Logic Quantum Enhancement Fix.toml

Eternalnacho:
Small extra condition to make Hazard cards compatible with Quantum enhancements - without, it obliterates max hand size.

(P.S. not sure why it's saying the match_indent line changed to itself, I didn't touch it)


InertSteak:
Makes sense, I'll merge it

PR #393, opened 2025-06-30
Should not need to replace ptype

RoyalRedRamble:
Applying the type sticker adds the sticker attribute and then also replaces the original type? That seems off, especially as so much type checking is doing the correct thing of checking for the sticker before the original type.

Replacing the original type is unneeded and removes useful information from the card.

I removed the following lines and spammed Tera orb, type sticker updates accordingly. Applying energy cards still works, transform energizes successfully and evolves the mon. Placed Grubbin alongside a tera orbed lighting mon and his ability still procced. 

@Eternalnacho was looking at reading the original type for an effect and we both ran into the offending code, it seems to go against what all the type_sticker code is trying to accomplish.

### DIFF
--- a/functions/pokefunctions.lua
+++ b/functions/pokefunctions.lua
@@ -918,10 +918,6 @@ apply_type_sticker = function(card, sticker_type)
     end
   end
   
-  if card.ability and card.ability.extra and type(card.ability.extra) == "table" and card.ability.extra.ptype then
-    card.ability.extra.ptype = apply_type 
-  end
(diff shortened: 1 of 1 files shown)


RoyalRedRamble:
Lol the branch name was supposed to be `ptype_trampling_fix`, oh well it's made now!

InertSteak:
Closing old pull requests. This was a workaround to another issue so I prefer to leave it in place for now

PR #394, opened 2025-07-01
Changed Nightmare's call to remove() to have "self" as the 1st parameter

Maelmc:
Even if currently unused, remove takes "self" as its first parameter and if it eventually gets used (for example, if self == card then we know it's a self destruct) then it's better than Nightmare passes "self" as the 1st parameter instead of "selected".


InertSteak:
Yeah good call

PR #424, opened 2025-07-07
JokerDisplay - Add Raikou, Entei, Lugia, and Ho-Oh. Update Staryu, Jigglypuff, Exeggcute

Definite-Internet-User:
I forgot the turn the pokemon discovery back on when I made a commit. I turned it back on in another commit though.


InertSteak:
Could you please send a commit that doesn't modify the config file?

Definite-Internet-User:
Does that fix it? If not I can do a git revert the 2 commits then redo the 2 commits without changing the config.

Edit: Also I can just redo the pr without modifying the config file in a new branch, close this pr and make a new pr if that's better for you.

InertSteak:
Yeah unfortunately it still shows the config file as being changed

PR #437, opened 2025-07-13
Japanese Sleeve localization

SonfiveTV:
I noticed that the Japanese localization for Sleeves uses スレーブ which sounds like Slave, instead of スリーブ which sounds like Sleeve

### DIFF
--- a/localization/ja.lua
+++ b/localization/ja.lua
@@ -4082,7 +4082,7 @@ return {
         },
         Sleeve = {
             sleeve_poke_pokemonsleeve = {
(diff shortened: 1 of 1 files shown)


khanryss:
The JP localization is lacking the last few updates because I haven't been playing much, and to be honest there's a lot that could be cleaned up / reworded when I (or if anyone else wants to take a stab at it!) get around to it.

That said, thank you for noticing this because I don't think I would have caught it! I think it's just a typo that then got copy pasted a bunch, hehe.

InertSteak:
Merging!

PR #438, opened 2025-07-13
Update ja.lua

khanryss:
Cleaning up some mistakes I made in the Nature tooltips.


khanryss:
I didn't actually update Sceptile's tooltip to reflect the updated effect, that will come in a more substantial cleanup.

PR #449, opened 2025-07-24
Added precision adaptation for mobile devices to enable usage on mobile platforms, and implemented a double-tap card feature to view evolutions on mobile devices. 

a2639443196:
Added precision adaptation for mobile devices to enable usage on mobile platforms, and implemented a double-tap card feature to view evolutions on mobile devices. huangjiarong 6 minutes ago


InertSteak:
Wow that's sick, nice job

PR #453, opened 2025-07-27
Shiny rival does not remove extra booster on self destruct

RoyalRedRamble:
As the title says:

https://github.com/user-attachments/assets/580fa73d-1242-4150-add8-75cb414b45e6

After the rival gets triggered and self destructs it wasn't removing the booster bonus.


InertSteak:
Good catch, thanks!

PR #454, opened 2025-07-27
Updated spanish translation (Everstone, config and minor fixes)

heyctf:
I have figured out yay
Minor fixes include just text placement, uppercase to lowercase for some terms and changed the name of Lightning type to be more faithful to TCG


InertSteak:
Nicely done!

PR #455, opened 2025-07-29
JokerDisplay - Fix massive bug in Mega Pidgeot

Definite-Internet-User:
I've heard that goto statements were the worst, I now see why. I used goto statements in my first try to update the pidgey line to the new effect but decided to do something different and apparently missed a remnant of that first try. Sorry about introducing a worse bug in a bug fix commit.

The dangling goto statement seemed to have made every definition in definitions1.lua non-functional or at least it made the charmander, krabby, bulbasuar, and voltorb lines non-functional.


InertSteak:
image

Lol no worries, happens to the best of us

PR #456, opened 2025-07-29
Fix mod crashing on case sensitive file systems

gsegalini:
After e54fb0f, some files for the clubs suite were renamed to have a capital S for skin, while the filename was not updated in pokesprites.lua.

This does not cause any issues on case-insensitive file systems like NTFS, but on other filesystems such as ext4 used on Linux the files are not found, crashing the game on boot.


InertSteak:
Good to know, thanks!

PR #458, opened 2025-08-01
Fix Evo Item Popups with Custom Content

LegendDuoKeeper:
The new addition to Family popups would crash when pulling up a non-native pokemon that uses a native evo item to evolve. (or vice versa if that made sense) I've reworked the bottom of the get_family_keys function a bunch to allow the separation of the pokemon's mod prefix from the evo item's mod prefix. 

I ran tests with every relevant combination of native and non-native elements I could think of, including branching evolutions having at least one item from each source. Despite finding no issues during my testing, I was still guessing on parts of this solution by the end, and would advise some additional testing before fully merging.

There's also certainly some room to clean up my solution to separating native and non-native evo items, but I'm pretty sure said improvements require things I do not yet know how to do.


InertSteak:
Wow, thanks for saving me the work on that. Good stuff

PR #459, opened 2025-08-02
Small quick fix (missing text for detailed info)

heyctf:
Sorry I forgot about this text, maybe I got confused with the title for the tooltip. That should be everything for the config.
I would also thank a lot if the text for the buttons could be localized, but for the spanish translation that's everything for now
image


InertSteak:
No worries

PR #468, opened 2025-08-11
Updated spanish translation for the contest winners!

heyctf:
I like the new jokers ! I finished making a translation for their descriptions, so yeah
I also changed some words to lowercase, I'm honestly not sure I left them with uppercase at the start but I fixed it.


heyctf:
There is a single space on a line on Lopunny's description that shouldn't be there...

InertSteak:
Nicely done!

PR #470, opened 2025-08-19
JokerDisplay - Add Contest winners and update Jolteon

Definite-Internet-User:
Getting Mega Lopunny to not crash upon leaving a blind made me want to tear my hair out.


InertSteak:
I can only imagine lol, thanks

PR #475, opened 2025-09-01
JokerDisplay - Fix bug with probability pokemon. 

Definite-Internet-User:


InertSteak:
Thanks, I was about to look into this but that saved me the trouble!

PR #484, opened 2025-09-07
Added soul_pos for all future Legendary Jokers

SonfiveTV:


InertSteak:
*Excellent*. My future self thanks you.

PR #490, opened 2025-09-19
Miscellaneous sprite modifications

Catzzadilla:
- Added an item card template for those who want to use it as a base for sprites
- Changed Arceus's Master Ball to Origin Ball
- Fixed Mega Tyranitar's belly being detached from the rest of its body, leading to possible sprite oversights

If you don't want some of these changes I think you can veto them and pull the rest, don't know for sure though.


InertSteak:
Good stuff as usual!

PR #493, opened 2025-09-25
Flipped the sprite on Linking Cord for visual aid

Catzzadilla:
Forgot to check in this one earlier. Left card goes up, right card goes down, right? This should help players remember it!


InertSteak:
Ah I see, good thinking!

PR #494, opened 2025-09-26
reverted changes to Mega Tyranitar per MyDude's request

Catzzadilla:
we shall never deny a guest 🦀


InertSteak:
even the most ridiculous request

PR #499, opened 2025-09-29
Added RedHaering's Lotad and Gothorita

Catzzadilla:
Lotad reveal will be awesome once we get into gen 3, will ask MyDude if he still wants to do the dancing Ludicolo too


InertSteak:
Neat!

PR #500, opened 2025-10-02
Removed dark pixels on MyDude's sprites

Catzzadilla:
they should look a lot better now in editions like foil and polychrome. Also Gengar is still hidden by its background, but its outline and shape can now be better seen if you drag the card in front of a bright object like the options button


InertSteak:
Excellent

PR #511, opened 2025-10-15
Slight correction to get_previous_evo()

Eternalnacho:


InertSteak:
It seems like this change also has your pending pre evo changes too (That's huge btw, thanks). 
Do you want me to just change that line myself?

Eternalnacho:
Either way works - if the pre-evo thing isn't quite what you want yet then by all means feel free to do just the one line

InertSteak:
Oh we very much want the pre-evo thing, just we want to make sure of things before we bring it in, since its a pretty big feature. I think it will be added soon though, and your effort is very much appreciated!

InertSteak:
Closing this as I already fixed it

PR #512, opened 2025-10-15
Joker Pre-Evos can get Stake Stickers

Eternalnacho:
I made a function (and corresponding config toggle) to allow the pre-evos of jokers to count as winning the run along with their evolutions


InertSteak:
After testing this, wow it works perfectly. Excellent work on this, saves me a ton of effort. Thanks and you're a total boss for this

PR #521, opened 2025-10-27
Joker_Family_Win fix for Gourgeist + Dudunsparce

Eternalnacho:
Just adding in a check to make sure that it grabs just the key and not the whole table for jokers that have table entries in pokermon.family


InertSteak:
Oh nice, beat me to the fix

PR #530, opened 2025-11-07
Made the mega sleeve shine

Catzzadilla:
used the negative_shine shader as it doesn't move too quickly and isn't too erratic for an object of this size, unlike the booster pack shader. Shout out to Emma for helping me with this


InertSteak:
Sweet!

PR #533, opened 2025-11-10
Implemented G.GAME.dollar_buffer

SonfiveTV:
This is something Vanilla Jokers use and have implemented it where necessary

I have tested and as far as players are aware, there shouldn't be any changes to the experience of these Jokers, however they now allow referencing the current money balance in the middle of a hand using (G.GAME.dollars + G.GAME.dollar_buffer)


InertSteak:
That is good to know! Thanks

PR #535, opened 2025-11-15
Art credits page

emmadenslemma:
Admittedly this should've been about 7-8 separate PRs instead, but here it is
(A lot of files have changes listed as the entire file removed and re-added, which is my editor changing the EoL sequence from CRLF to LF. Sorry about that)

### pokespriteload.lua

- `PokemonSprites` is now an ordered list with a lookup table attached. Indexing it works the same as usual thanks to metatables.
- `gen_atlas` and `lookup_gen_atlas` are now numbers instead of booleans. This lets you separate sprites from generation toggles (I.e. have Alolan Diglett disappear with Gen 7 turned off) but more importantly lets me find the atlases for sprites that do not have jokers implemented yet
- All of the legendaries, regionals, pre-ZA megas, and GMAX pokemon (should) have gotten entries, with some forms left out due to ambiguity of implementation
- Artists in the `alts` field can now be a table instead of a string. The way it differentiates between these and lists of artists is whether or not the `name` field is present. This is currently only used to add the `layer` field, which is set `center` for full art background credits, and `soul` for floating sprite credits (All the requisite full art credits have been changed).
- Artists in the `poke_artist_info` table can now have a `links` field, which is a list of tables containing the fields `account`, `site`, `url`, and `colour`. Every field except `url` is optional and will be pulled from the url if missing.
- and lastly atlas concatenation has been moved to a separate function `poke_get_atlas_string`

### uifunctions.lua

- New helper functions `poke_create_UIBox_your_collection` and `poke_UIBox_link_button` have been added, with the former worked into both the Pokedex UI code and Sprite Change UI

### inputmanager.lua

- Basic Observer pattern input manager singleton to consolidate all of the right-click and double-click code into one place. Replaces double-click patches for opening the Pokedex UI and right-click patches for opening the Pokedex UI and using the Toggle Individual Sprites UI

### displaycard.lua

- New class used to display sprites without needing to have game objects exist.

### pokeui.lua

- The Config tab has been re-arranged to have all of the functions together, with some unused toggles removed. Functionality should be completely untouched
- Actual Credits tab has been moved away from hardcoding all the UI-elements to generating them programmatically, likewise functionally nothing should be different about it
- Art Credits page has been added!
- Pokedex UI has been rewritten to use `poke_create_UIBox_your_collection` and the input manager from earlier
- Toggle Individual Sprites UI has been `poke_create_UIBox_your_collection` and the input manager from earlier
- `poke_artist_credits` has been changed to do a lookup on the artist value to get the colours, removing the need to add fields beyond `artist` to cards, and also changing the `artist` field from a display name to an actual key in the `poke_artist_info` table (This has also been retroactively changed on all of the artist fields)

and that should be pretty much everything.
I've doubtless missed some stuff so @ me if you have questions about the code, when you need me to split it into the requisite 7-8 PRs, or when you inevitably discover a myriad of crashes by testing it, and I'll do my best to patch it up

### DIFF
--- a/backs/backs1.lua
+++ b/backs/backs1.lua
@@ -81,6 +81,7 @@ local revenantdeck = {
 	order = 26,
   unlocked = true,
   discovered = true,
+  artist = "Sonfive",
 	config = {},
   loc_vars = function(self, info_queue, center)
     return {vars = {localize("silverseal_variable")}}
@@ -98,6 +99,7 @@ local telekineticdeck = {
 	order = 19,
   unlocked = true,
   discovered = true,
+  artist = "MyDude_YT",
 	config = {vouchers = { "v_crystal_ball"}, consumables = {'c_poke_twisted_spoon', 'c_poke_twisted_spoon'}},
   loc_vars = function(self, info_queue, center)
     return {vars = {localize{type = 'name_text', key = 'v_crystal_ball', set = 'Voucher'}, localize("twisted_spoon_variable")}}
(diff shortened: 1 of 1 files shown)


InertSteak:
Absolutely incredible, you're a super star

PR #539, opened 2025-11-17
Remove erroneous Master Ball from Celebi placeholder backgrounds (my fault)

CBMX-609:
Celebi had floating master balls on the placeholder backgrounds. This pull should fix that

oops_ball


InertSteak:
Good catch

PR #540, opened 2025-11-18
Added Placeholder Background for Ho-Oh

Catzzadilla:
Same deal as Moltres and Celebi. Fire Flying legendaries sure have it rough, eh?


InertSteak:
50% damage from stealth rock will do that

PR #542, opened 2025-11-20
Fix artist grid showing IDs instead of artist display names

emmadenslemma:
well this is embarrassing, I knew I missed something


InertSteak:
Happens to the best of us

PR #544, opened 2025-11-23
Telekinetic Sleeve Combo

Eternalnacho:
This leaves the telekinetic sleeve intact, but adds the interaction with telekinetic deck as outlined in the fan contributions page


InertSteak:
Very nice!

PR #546, opened 2025-11-23
add mega deck sleeve combo

emmadenslemma:
changing config values in `loc_vars` might seem wonky but that's what the Card Sleeves mod does with the vanilla sleeve combos.


InertSteak:
All good, I gave it a test and fixed a few issues and now it looks good!

PR #549, opened 2025-11-24
Fix multi artist credits

emmadenslemma:
Fixes Rotom Dex (and other basic series arts with multiple artists) not showing up in the art credits page, and changes the existing pokespriteload multi-artist support to use the same functionality

Also center-aligns artist links when additional rows of less than 3 links are present


InertSteak:
Awesome, thanks!

PR #552, opened 2025-11-26
Creating a "Pokermon Jokers in Collection Only" Toggle and Dex Ordering Add-On jokers

Eternalnacho:
This should be everything to get both fully functional but please do make sure everything looks as you would like it!


InertSteak:
Great work! This is really cool

PR #553, opened 2025-11-26
Dex Order Event Fix

Eternalnacho:
Whoops, I missed a spot - the event I was using to order everything in the main file wasn't returning anything like it needed to. I also threw in a check for existing dex_order_groups in case fan additions need to create custom groupings to get things to order correctly


InertSteak:
Ah good catch

PR #554, opened 2025-11-26
Dex Order Event Fix x2

Eternalnacho:
Sometimes I do silly things like putting important tables inside events when they shouldn't be.


InertSteak:
Don't we all?

PR #556, opened 2025-11-28
JokerDisplay partial update

Definite-Internet-User:
Added Joker Display support for
- Lucario
- Pawniard line
- Charcadet line
- Absol
- Duskull line
- Excadrill
- Kricketot line

So I was going to also add joker display support for some of the other recently added pokemon but I never got around to it so I figured I would make a belated PR for the pokemon I did add support for.


InertSteak:
Nice work! Thanks!

PR #557, opened 2025-11-30
Unown Swarm Sticker "fix"

Eternalnacho:
So um. This is so extremely cursed, but the stickers are in the right place...


InertSteak:
Cursed indeed but if it works it works

Could I ask you to move the DrawStep calls to pokedraw.lua? I want to keep things as organized as possible

Eternalnacho:
Think I got everything moved properly

InertSteak:
Color me impressed with this solution. I never would have gotten something like this working lol

PR #558, opened 2025-12-02
Mega Evolution tooltip changes

emmadenslemma:
Moves the 'Can Mega Evolve' tooltip into `type_tooltip', complete with showing the 'XY Mega' tooltip for pokemon with multiple Mega evolutions using localized joker names.


InertSteak:
Nice

PR #560, opened 2025-12-06
Joker Display Update

Definite-Internet-User:
Added joker display support for
- Rockruff line
- Ferrothorn
- Rotom & variants except RotomDex
- Pumpkaboo line

Updated/bug fixed joker display definitions for
- Yamper was using the mult value for both mult and money
- Boltmund, was previously displaying 0x when not making a straight
- Pikachu, was always displaying +$2 even when above 25 dollars. Reversed order to mult, money to be more in line with other joker display definitions in pokermon
- Raichu now also displays money gain
- Golett line updated to be more in line with current effect
- Updated Swampert to display number of tarots that would be generated
- Updated Blaziken to display its xmult effect when procced
- Treecko, joker display simplified now it doesn't have chance effect
- Celebi updated to show its new xmult effect
- Poliwag line, added suit appropiate color to reminder text
- Poliwrath bug fix, xmult calculation wasn't using the correct math
- Flareon bug fix, if first mult card was highlighted it would display 1x even if there was another mult card that wasn't highlighted
- Gigalith's joker display retrigger logic updated to apply to rankless cards instead of solely hazard cards.
- Marill and Azumarill now accounts for debuffed cards

### DIFF
--- a/jokerdisplay/definitions3.lua
+++ b/jokerdisplay/definitions3.lua
@@ -3,7 +3,7 @@ local jd_def = JokerDisplay.Definitions
 --	Treecko
 jd_def["j_poke_treecko"] = {
   text = {
-    { text = "$", colour = G.C.GOLD},
+    { text = "+$", colour = G.C.GOLD},
     { ref_table = "card.joker_display_values", ref_value = "money", colour = G.C.GOLD },
   },
   reminder_text = {


InertSteak:
Awesome work as always, thanks!

PR #561, opened 2025-12-06
Gen 2 roundup

Catzzadilla:
Added more custom art, and updated the README as well, since we've fully implemented gen 2 lol


InertSteak:
The readme update is clutch

PR #562, opened 2025-12-07
Megastone Changes / Rework for Ditto Copies etc.

Eternalnacho:
The functionality of the mega stone should be largely the same, but the get_mega function for jokers with >1 mega now is an isolated function. Also now ties the used_on variable to a specific joker, rather than a key, meaning things like ditto / copy effects will not allow the mega stone to be used on the copies.

(Please do check and make sure it works how it should, I tried to make the use cases exactly the same but things can always slip through the cracks)

### DIFF
--- a/consumables/mart1.lua
+++ b/consumables/mart1.lua
@@ -296,6 +296,14 @@ local transformation = {
   end
 }
 
+local get_mega_target = function(self)
+  return poke_find_leftmost_or_highlighted(function(joker)
+    return not self.ability.extra.used_on and get_mega(joker) and not joker.debuff
+        or joker.config.center.rarity == "poke_mega" and joker.unique_val == self.ability.extra.used_on
+        or G.GAME.modifiers.infinite_megastone and ((get_mega(joker) and not joker.debuff) or joker.config.center.rarity == "poke_mega")
+  end)
+end
+
 local megastone = {
   name = "megastone",
   key = "megastone",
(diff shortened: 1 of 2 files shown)


InertSteak:
Unfortunately this doesn't seem to work correctly, at least on my end. I am unable to actually mega evolve anything for some reason. Looks like the target variable is always nil in the can_use function.

consumables/mart1.lua
```diff
@@ -321,103 +324,67 @@ local megastone = {
...
-          target = poke
-          break
-        end
-      end
-    end
+    -- Find an eligible pokemon (also checks if the mega stone has been used, and on which joker)
+    local target = poke_find_leftmost_or_highlighted(function(joker)
+      return not card.ability.extra.used_on and get_mega(joker) and not joker.debuff
+        or joker.config.center.rarity == "poke_mega" and joker.unique_val == card.ability.extra.used_on
+        or G.GAME.modifiers.infinite_megastone and ((get_mega(joker) and not joker.debuff) or joker.config.center.rarity == "poke_mega")
```
emmadenslemma: mixing `and`s and `or`s should use parentheses for clarity
(ex. `(a and b) or (b and c) or (c and d)`

consumables/mart1.lua
```diff
@@ -321,103 +324,67 @@ local megastone = {
...
-      for k, poke in pairs(G.jokers.cards) do
-        if (poke.config.center.megas or poke.config.center.rarity == "poke_mega") and not (poke.debuff and poke.config.center.megas) then
-          target = poke
-          break
-        end
-      end
-    end
+    -- Find an eligible pokemon (also checks if the mega stone has been used, and on which joker)
+    local target = poke_find_leftmost_or_highlighted(function(joker)
+      return not card.ability.extra.used_on and get_mega(joker) and not joker.debuff
```
emmadenslemma: `get_mega` takes a center not a card object

functions/pokefunctions.lua
```diff
@@ -672,6 +672,17 @@ get_highest_evo = function(card)
   return pseudorandom_element(evos, pseudoseed('highest'))
 end
 
+get_mega = function(card)
```
emmadenslemma: change parameter `card` to `center` for clarity,
or add `local center = card.config.center` inside of `get_mega`

Eternalnacho:
I assume this is the case, but that get_mega function was in pokefunctions when you tested it right? that's the only reason I'd see it coming up nil _every_ time

Eternalnacho:
okay nevermind should be I got all the errors (also tysm emma, enormously helpful)

InertSteak:
Ok, I'll test again when I get the chance. Thanks for the hard work!

InertSteak:
Looks good now, awesome work on this!

PR #563, opened 2025-12-07
Luminous deck patch fix

Eternalnacho:
I missed a spot in the energy refactor - energy_increase needs to be called with get_type(card) instead of type_sticker_applied


InertSteak:
Good catch!

PR #564, opened 2025-12-09
Added Unown Full art Background by Catzzadilla, as well as Title Card sprites for the main menu

Catzzadilla:
This should make the R in pokermon more readable. I also made the Basic series Unown background negative colored, so it might be a good idea to make Unown always spawn negative in the collection and such.


Catzzadilla:
also added a disclaimer under Moltres, Ho-Oh and Celebi's Series A arts in-game that says that these sprites are currently placeholders.

PR #566, opened 2025-12-09
Adjusted Anorith card creation logic

SonfiveTV:
Copied the logic from Bonsly, it works now with Jynx


InertSteak:
Thanks for the effort, this would have worked. However, I instead added poke_add_playing_card as a function to both call SMODS.add_card and call the playing card effects function. This should be used from now on for these effects. Probably need to refactor some old mons later

PR #567, opened 2025-12-10
Zorua shop illusion fix

Eternalnacho:
what it says on the tin, ensuring the sprites are set after the card properly has an area


InertSteak:
Awesome, thanks!

PR #571, opened 2025-12-12
standardize evo form handling

emmadenslemma:
changes the Gourgeist exception in `poke_backend_evolve` to work with any evolving mons that use forms

(Shouldn't change any functionality since Gourgeist already duplicates this inside `set_ability`)


InertSteak:
Awesome, thanks!

PR #573, opened 2025-12-17
Updated sprites

Catzzadilla:
- Tatsugiris on the base atlas have different expressions (thanks Emma!)
- Updated CBMX's Ditto


InertSteak:
Nice

PR #574, opened 2025-12-17
Replaced Chartreuse Chamber with Team Star

Catzzadilla:
- Made Penny's glasses lighter
- Replaced Chartreuse Chamber with Iridescent Hacker (nothing changed functionally, just name, colour, texture, etc.)
- Implemented The Star, which is a "lite" version of Chartreuse Chamber that will only ever hit 1 type


InertSteak:
Looks good!

PR #576, opened 2025-12-18
vanilla joker energy fixes

Eternalnacho:
mostly just jokers that should be xmult_multi that used to be erroneously classified as xmult_mod, but also Canio being fixed to reflect vanilla Balatro calling it 'Caino' in the code


InertSteak:
Looks good!

PR #578, opened 2025-12-19
fix crash when mirror transforms a form joker

emmadenslemma:
Quick null check because I somehow forgot that you can turn form jokers into other jokers, and that `new_card` is a center and not a card


InertSteak:
Oh good catch

PR #579, opened 2025-12-19
Smeargle function relocation and slight rework

Eternalnacho:
Most of what this does is move functionality out of Update and into Calculate - it also makes the load and add_to_deck functions unnecessary. As always, please do check and make sure nothing breaks if interested


InertSteak:
I'm impressed, I wasn't able to break it. I wanted to cause a stack overflow too, those are always fun

PR #580, opened 2025-12-22
make oinkologne multisprite

emmadenslemma:
small PokemonSprites change
(yes this is technically a fan addition change for Sonfive's Pokermon+)


InertSteak:
Seems fine to me

PR #583, opened 2025-12-26
increment_energy wasn't properly working with energy_types

Eternalnacho:
Seems I missed a spot, just changes the energize function call so etype is used correctly


InertSteak:
Looks good!

PR #584, opened 2025-12-27
artist & designer credits on blinds in collection

Maelmc:


InertSteak:
Excellent! Looks good to me

PR #585, opened 2025-12-27
Pocket Pack Cleanup

emmadenslemma:
Changes the evo item generation code of `create_item`, using `G.GAME.used_jokers` to overcome the need for tracking created items, and using the (now moved) `native_evo_items` list to support fan addition evo items without patches

Functionality should be identical for all base mod evo items

Additionally I also added `.vscode/` to the gitignore file, for local workspace settings (I grew tired of staring at yellow lines)


InertSteak:
Great stuff

PR #586, opened 2025-12-28
get_highest_evo fixes

Eternalnacho:
Saw a bug report about Hitmonlee evolving to Hitmontop via transformation, this _should_ fix any such errors alongside properly checking for the (admittedly very small) edge case of a split evo with different stage amounts (hydrapple, mr. rime)


InertSteak:
Yeah I saw that too. Doing Arceus's work Nacho!

PR #587, opened 2025-12-30
Consumable text is more consistent

Catzzadilla:
- Changed "o" on Thunder Stone and linking cord to match Twisted Spoon, Microwave, and bold letters on other cards
- Changed "P" on Prism scale to match Twisted Spoon
- Changed "J" on Berry Juice to be more "J" and less "⅃"
- Removed weird shading on and around Leftovers's text


InertSteak:
Nicely done

PR #588, opened 2025-12-30
changing get_family_keys to only require one argument

Eternalnacho:
also sneaking in a probably redundant but still helpful null_check for is_type() juuuuust in case.


InertSteak:
Awesome, thanks!

PR #592, opened 2025-12-31
Tyranitar Series A ultra ball fix

Catzzadilla:
saw that it was too dark while working on a new thumbnail for Pokermon in BMM, lol


InertSteak:
Ah I see good catch

PR #593, opened 2025-12-31
add `saveable` option to consumables

emmadenslemma:
Changes the hardcoded exception for Mega Stone to always be saveable from Spectral packs, to a generic `saveable` field for reusability


InertSteak:
Oh nice, good thinking.

PR #596, opened 2026-01-01
Moved Nature definition of Nature jokers to their own function

Maelmc:
This would allow better Nature manipulation from external sources (ex: nature mints that'd reroll the nature). You can't just call `set_ability` because:
- it's not `initial` (though you could just set `initial` in the function call)
- Specifically Unown does a bunch of other things in its `set_ability` function, and so calling it would for example change its form, and other future jokers with a Nature could do other things in this function than setting a nature too

Also, that means you can check if a Joker has a `set_nature` function to know if it has a nature (I guess that already worked by checking if it had a `targets` field in its ability, but `targets` in general can mean other things that "this joker has a nature")

Your call if you wanna implement that. Currently it doesn't do much, but I think it's better to do it now that there's only 15, than in the future when we have many more Nature jokers and suddenly we need to do something similar for all of them because of a joker or consumable concept

### DIFF
--- a/pokemon/pokejokers_07.lua
+++ b/pokemon/pokejokers_07.lua
@@ -1097,13 +1097,16 @@ local unown={
       card.ability.extra.form = form
       self:set_sprites(card)
       
-      card.ability.extra.targets = get_poke_target_card_ranks("unown", 1, card.ability.extra.targets)
+      self:set_nature(card)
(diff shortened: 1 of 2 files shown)


InertSteak:
Yeah this seems like a good idea.

Issue #599, opened 2026-01-03
[SMODS Pokermon "functions/energyfunctions.lua"]:177: attempt to perform arithmetic on a table value


KassLavender:
dragon sticker on a normal fibonacci joker, using a dragon energy made balatro not likey
perhaps could use a pcall? sorry my modlist is embarassingly large
```
[SMODS Pokermon "functions/energyfunctions.lua"]:177: attempt to perform arithmetic on a table value
Stack Traceback
===============
(1) Lua upvalue 'orig' at file 'main.lua:612'
	Local variables:
	 msg = string: "[SMODS Pokermon \"functions/energyfunctions.lua\"]:177: attempt to perform arithmetic on a table value"
	 (*temporary) = Lua function '?' (defined at line 31 of chunk [SMODS _ "src/logging.lua"])
	 (*temporary) = string: "Oops! The game crashed\
"
(2) Lua local 'handler' at file 'debugplus/console.lua:607' (from lovely module debugplus.console)
	Local variables:
	 msg = string: "[SMODS Pokermon \"functions/energyfunctions.lua\"]:177: attempt to perform arithmetic on a table value"
(3) LÖVE metamethod at file 'boot.lua:352'
	Local variables:
	 errhand = Lua function '(LÖVE Function)' (defined at line 606 of chunk [lovely debugplus.console "debugplus/console.lua"])
	 handler = Lua function '(LÖVE Function)' (defined at line 606 of chunk [lovely debugplus.console "debugplus/console.lua"])
(4) Lua global 'energize_other' at file 'functions/energyfunctions.lua:177' (from mod with id Pokermon)
	Local variables:
	 card = table: 0x10c304c0  {click_offset:table: 0x10c30868, static_rotation:false, ambient_tilt:0.2, parent:table: 0x171622e8, T:table: 0x10c306b0, last_aligned:-1, extra_cost:3 (more...)}
	 etype = string: "Dragon"
	 center = table: 0x0a1b1b28  {loc_vars:function: 0x0a59b8f0, unlocked:true, registered:true, _discovered_unlocked_overwritten:true, order:31, _d:false, set:Joker, set_card_type_badge:false (more...)}
	 colorless_penalty = number: 1
	 amount = number: 1
	 field = string: "mult_mod"
	 value = string: "extra"
	 config = table: 0x09aed5b8  {extra:table: 0x09aed5e0}
[...]

PR #600, opened 2026-01-03
energyfunctions.lua:energize_other compatibility with Minty

KassLavender:
[Minty](https://github.com/wingedcatgirl/MintysSillyMod) changes how some particular vanilla jokers hold their config stats; ex. for Fibonacci, ```config = {extra = 8}}``` becomes ```config = {extra = {mult = 8, again = 0}}}```. This PR fixes crashes involved in energizing particular values that are expected to be numbers but have become tables with the relevant number inside, and maintains the intended functionality with and without Minty. See [this pokermon issue](https://github.com/InertSteak/Pokermon/issues/599) for the discussion


InertSteak:
Makes sense, I'll merge this

PR #601, opened 2026-01-03
Fix Joker Pool interactions

emmadenslemma:
Fixes an issue where opening the Pokedex View outside of the collection adds the family members to the list of current jokers for the run, removing them from the pool

Also fixes an issue where using a Mega Stone in the booster pack instead of saving it doesn't remove it from the pool


InertSteak:
Looks good!

PR #607, opened 2026-01-05
fixing crash with energizing vanilla jokers

Eternalnacho:
Variable check for the value type "number" needed to be in quotes, was always returning false


InertSteak:
Ah I see, didn't catch that

PR #609, opened 2026-01-06
Minor matching_energy() optimization

Eternalnacho:
Noticed that matching_energy wasn't taking advantage of get_type(), and that made the code for it a lot simpler
Also made poketype_list a global variable so that anything that needs to reference it can without making a new list every time.


InertSteak:
Ah I see, looks great!

PR #610, opened 2026-01-06
Fix artist credit tooltip function

emmadenslemma:
Small refactor to fix a bug where highlight colours of a second artist would be applied to the first instead.

Also re-indents the function for readability


InertSteak:
Looks good, thanks!

PR #611, opened 2026-01-07
fix `saveable` field

emmadenslemma:
I must've been very distracted while adding it the first time, because I seemingly only added it to the Controller UI
This should work for both and also de-duplicate the check for when you can save consumables


InertSteak:
Happens to the best of us, looks good.

PR #613, opened 2026-01-08
Family Optimizations

emmadenslemma:
Moves the family list and corresponding functions to their own file (bar get_family_keys which has enough adjacent functionality to warrant staying put) and changes the family list to a map instead, with each key holding a reference to its own family list

(should also be tested by someone other than me for such a core part of the mod)


emmadenslemma:
Currently has an issue adding forms via addons

emmadenslemma:
I have ironed out all of the issues that I could find

InertSteak:
After testing a bunch, I couldn't really break anything. So I think this looks good. We can clear up any issues as they crop up. Big win for optimization!

PR #617, opened 2026-01-12
one more dex order fix

Eternalnacho:
missed a spot, needed this function to look for a name instead of a key


InertSteak:
Ah I see gotcha

PR #623, opened 2026-01-15
Add Bitter Rival to Art Credits

emmadenslemma:
Or more accurately: Add support for credits on basic series sprites.

Still not the preferred way of doing this for new sprites; pokespriteload needs the display text shown in the credits passed manually, since it has no relation to the jokers (since they aren't guaranteed to exist).

Additionally the fallback to look for credits defined on the jokers still exist, albeit now it will completely disregard those if **any** credits for the card are defined in pokespriteload, **regardless of the current series being used**.


InertSteak:
Nicely done

PR #624, opened 2026-01-15
Added art credits to Other Jokers

Catzzadilla:
really thought we had professor in there but I guess not


InertSteak:
Ah I see, well good to get those in now

PR #626, opened 2026-01-16
refactor professor in_pool method

emmadenslemma:
one small self-inflicted communication error later, we arrive

note that if you test this working, spawning jokers via debugplus does not make them "present" as far as the family check is confirmed. (troubadour is unsurprisingly helpful here)


InertSteak:
Looks good! and yes troubadour is a godsend honestly

PR #627, opened 2026-01-16
Bug Fix for Dex Order + Joke Content

Eternalnacho:
adding a default case to get_dex_number to avoid giving the billion lions an advantage (crash fix)


InertSteak:
Lol,

PR #628, opened 2026-01-16
make the main menu displaycard unselectable on controller

emmadenslemma:
fixes an issue where the main menu card could be selected through settings menus


InertSteak:
Oh awesome, I wasn't sure how to fix that one

PR #629, opened 2026-01-17
Add Tag and Seal credits

emmadenslemma:
Changes the following:

- Tags now display Artist and Designer credits
- Seals now display Artist and Designer credits in the collection only
- Multiple artists are no longer displayed "A & B" but instead "A and B" with 2 artists, and "A, B, and C" with 3 or more artists. (uses the localization key `poke_and`)


InertSteak:
Excellent work

PR #633, opened 2026-01-19
Changed Relicanth to Earth Type

Catzzadilla:
thanks Maelmc for pointing it out
it should now synergize better with Hard Stone


InertSteak:
Looks great!

PR #634, opened 2026-01-19
Challenge button colours

Catzzadilla:
ngl this is so sick. Had to put a lovely patch in there until smods pulls a request a year from now, but this works and omg do the challenges actually look pretty

next step is making the banned blinds not take up the entire screen when you have a lot of them lol


InertSteak:
Sweeeet

PR #635, opened 2026-01-19
Let a Pokémon be discovered if it's set to be discovered from base

Maelmc:
Currently this does nothing for Pokermon, except remove that Unown exception in pokemon_load, but it'd let you have more jokers discovered by default more easily in the future if you want.

For me though, I have some multiplayer Jokers that load regardless of if you have Multiplayer or not (to show that I have some multiplayer content), so I need them to be discovered by default.


InertSteak:
Ah I see cool, thanks!

PR #636, opened 2026-01-20
reworking the elemental monkeys (and probopass, smeargle and aipom)

Eternalnacho:
This eliminates 'scuffed_af', as well as the long lovely patch associated with the elemental monkeys and their respective joker effects, and replaces them with function hooks where applicable (and one much smaller lovely patch)


InertSteak:
Oh awesome! Rip to the scuffed_af though, you will be missed

PR #637, opened 2026-01-20
Cleanup file loading

emmadenslemma:
Shortens the file loading code in the main pokermon.lua file

Also crashes early if any of the helper function files are missing, as a treat


InertSteak:
Love to see the cleanup

PR #640, opened 2026-01-21
Fix Unown Series A credits

emmadenslemma:
Properly adds LocalThunk to the Unown Series A sprite credits, and fixes a bug where layer credits for artists that aren't in the artist info table would crash the game


InertSteak:
Nice

PR #644, opened 2026-01-22
Move Bidoof/Bibarel English localization

emmadenslemma:
Moves the default localization for Bidoof/Bibarel out of the joker declarations to give localizers a better overview


InertSteak:
Thank you! I totally forgot to do that

PR #645, opened 2026-01-23
Fix wrong variable being set in `load_pokemon`

emmadenslemma:
Small typo fix, shouldn't change any current functionality as all of the current evo_list pokemon (gloom, poliwhirl, and charcadet) have their `config.extra` tables defined


InertSteak:
Looks good

PR #647, opened 2026-01-24
Fix flower card animation

emmadenslemma:
To commemorate Gunmetal Guy making the flower card sprites, I decided to hop in and fix all the issues surrounding timing and sprite application with these ones

As usual more testing than I've done is probably required, I've only tested animations with 1X speed, and interactions with Midas Mask and Nosepass

Also some of these could be personal preference, such as the final step doing the sound-effect, popup text, sprite change, and money clink at the same time. Let me know if any of that needs changing


InertSteak:
Wow it looks great now, thanks!

PR #648, opened 2026-01-24
Added Seed and Flower card art by MyDude (WIP)

Catzzadilla:
There are still some changes to be made, but it's coming to fruition more and more!

Also removed the patch that added custom challenge button text colours, since it was added in the latest Steamodded update


InertSteak:
Wicked sick

PR #649, opened 2026-01-24
add temporary fix for bramblin visuals

emmadenslemma:
Not ideal, but a good enough holdover until I fix it properly


InertSteak:
Good enough for now haha

PR #650, opened 2026-01-24
Properly fix Bramblin visuals

emmadenslemma:
Should be working smoothly now, and also require less outside intervention when creating Seed cards
Also fixes Roselia's visuals


InertSteak:
Awesome stuff and quick work! Thanks!

PR #652, opened 2026-01-26
Editions on soul layers

emmadenslemma:
Adds a toggle to force editions to apply to the soul layer, for use with custom sprites where the soul layer and the center are connected

Feature in action:
image


InertSteak:
Wow that's sick

PR #654, opened 2026-01-27
Added PurpleZaffre to the art credits page.

Catzzadilla:
PurpleZaffre doesn't have any sprites in the mod as of yet, but they've made most of the Evil Team Leader Boss Blind sprites in the atlas. I reached out to them asking them for permission to use these sprites, and they said yes!


InertSteak:
Nice!

PR #656, opened 2026-01-28
Add `pokeconstants.lua`

emmadenslemma:
Admittedly a smaller refactor than I'd like as it doesn't include extended_family or the energy constants, since the energy constants sit comfortably inside energy functions and extended_family needs a bigger refactor than this, but if nothing else this can still be a guiding ruler for where to put tables like these in the future

includes aliases to any globals I've renamed, both for compat with fan additions, and also as to not redo the entire codebase changing `poketype_list` to `POKE_TYPES` 40 different places, so functionality should be completely unaffected by this

### DIFF
--- a/functions/pokefunctions.lua
+++ b/functions/pokefunctions.lua
@@ -401,33 +401,6 @@ deck_seal_evo = function (self, card, context, forced_key, seal, percentage, fla
   end
 end
 
-POKE_STAGES = {
-  ["Baby"] = { prev = nil, next = "Basic" },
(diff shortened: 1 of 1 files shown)


InertSteak:
Gonna wait on this, since I'm going to make another release version soon. But as soon as I do, I'll merge this. Stellar work as always

InertSteak:
Accidentally added the constants back to pokefunctions, but I fixed it. I'll do some testing, but this looks good!

InertSteak:
Ok, I wasn't able to break it. We'll deal with any issues as they pop up

PR #658, opened 2026-01-29
Changed challenge colour to be more uniform + PurpleZaffre Crash Fix

Catzzadilla:
I didn't know this, but apparently challenges in a mod typically all share the same colour so it's easy to identify at a glance what mod the challenge comes from. I originally wanted the colours to reflect what the challenge was about, but it might be better to have something like little icons for that instead in the future. Some were hard to read for people too. I recolored them to the colors of the Pokémon logo, same as the streaks of colour in Pokermon's own logo.


InertSteak:
Yeah it was a cool idea, just needs some refinement

PR #659, opened 2026-01-29
Evo item gen fix

emmadenslemma:
fixes a bug where pocket packs wouldn't try to force evolution items if you only had 1 joker that evolved by item, and hopefully fixes a crash with Multiplayer Small World (unable to test this)


InertSteak:
Ah yeah, good catch!

PR #660, opened 2026-01-29
adding talisman compatibility to brambleghast

Eternalnacho:
Was in a run and crashed with Brambleghast, but it's just a Talisman thing, so this should hopefully fix that


InertSteak:
Damn, somehow missed this PR. My bad. Let me get this merged finally lol

PR #661, opened 2026-01-29
adjusting energy count logic

Eternalnacho:
Apparently somewhere in my code delirium I made a weird logic choice, so energy_count and c_energy_count weren't incrementing correctly. Whoopsie.


InertSteak:
Oops, meant to merge this earlier. Looks good

Issue #630, opened 2026-01-17
pokemon crashes on startup (android)


Fuddung:
Used to work two weeks ago, now just crashes

Oops! The game crashed:
[SMODS _ "src/game_object.lua"]:433: Failed to collect file data for Atlas poke_logo_alt

Additional Context:
Balatro Version: 1.0.1o-FULL
Modded Version: 1.0.0~BETA-1224a-STEAMODDED
LÖVE Version: 11.5.0
Lovely Version: 0.8.0
Platform: Android
Steamodded Mods:
    1: Pokermon by InertSteak, Please See Credits Page [ID: Pokermon, Priority: -1, Version: 3.6.0-0113a, Uses Lovely]
Lovely Mods:

Stack Traceback
===============
(3) global C function 'assert'
(4) Lua method 'inject' at Steamodded file 'src/game_object.lua:433' 
Local variables:
 self = table: 0x78944f9338  {mod:table: 0x799e20f5f8, full_path:Mods/Pokermon-main/assets/2x/poke_logo_alt.png, px:389, prefix_config:table: 0x78944f9380, name:poke_logo_alt (more...)}
 file_path = string: "poke_logo_alt.png"
(5) Lua method 'inject_class' at Steamodded file 'src/game_object.lua:183' 
Local variables:
 self = table: 0x79a5b35db8  {__call:function: 0x79a5b6d690, subclasses:table: 0x79a5b35e50, set:Atlas, obj_buffer:table: 0x79a5b35948, super:table: 0x79a5b34100, __index:table: 0x79a5b35db8 (more...)}
 inject_time = number: 1.08039
 start_time = number: 6.50008
 end_time = number: 6.50008
 o = table: 0x78944f9338  {mod:table: 0x799e20f5f8, full_path:Mods/Pokermon-main/assets/2x/poke_logo_alt.png, px:389, prefix_config:table: 0x78944f9380, name:poke_logo_alt (more...)}
 (for generator) = C function: builtin#6
 (for state) = table: 0x79a5b35948  {1:mod_tags, 2:achievements, 3:poke_j_poke_unown_swarm, 4:poke_j_poke_unown_swarm_shiny, 5:poke_j_poke_unown_swarm_soul, 6:poke_j_poke_unown_swarm_shiny_soul (more...)}
 (for control) = number: 84
 i = number: 84
 key = string: "poke_logo_alt"
(6) Lua field 'injectObjects' at Steamodded file 'src/game_object.lua:257' 
[...]

PR #663, opened 2026-02-05
Fix for issue 630

JTBringe:
Balatro crashes on mobile because it doesn't recognize the file "poke_logo_alt.png".
Actual filename was "Poke_logo_alt.png", and Atlas is apparently case sensitive 😀
This didn't cause issues on desktop, which is odd, only on mobile.

Changed first character to lowercase p for "Poke_logo_alt.png" in both 1x and 2x folders, which fixed the crash reported in Issue 630: https://github.com/InertSteak/Pokermon/issues/630

Thanks to [DeezIsMyUsername](https://github.com/DeezIsMyUsername) for outlining the fix, I just did the PR :D


InertSteak:
Seems good to me

PR #668, opened 2026-02-10
Refactor Scaling Jokers (Gen 1)

emmadenslemma:
First of many refactors to make everything compatible with the `SMODS.scale_card` API.
Tested on BETA 1224a but given that scale_card hasn't received any changes for ~400 version numbers it should work on 1016c :pray: 
Additionally fixes:
- Parasect letting you go below +0 Mult
- Crobat working with some quantum enhancement combinations but not others
- Aerodactyl and Gholdengo keeping their increased starting XMult when energized
- Slowking and Mega Houndoom keeping their increased starting XMult when energized during a Blind

Note that Crobat will give you stat boosts for *every* Enhancement you have on a card, quantum or otherwise, when triggered. Let me know if Crobat working with Quantum Enhancements is a bug and I will change it.

### DIFF
--- a/pokemon/pokejokers_05.lua
+++ b/pokemon/pokejokers_05.lua
@@ -1379,17 +1379,17 @@ local dratini={
   perishable_compat = false,
   blueprint_compat = true,
   calculate = function(self, card, context)
-    if context.cardarea == G.jokers and context.scoring_hand and context.full_hand then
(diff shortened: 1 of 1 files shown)


InertSteak:
Amazing work as always. Thanks a bunch!

PR #670, opened 2026-02-11
Refactor Scaling Jokers (Gen 3-4-5)

emmadenslemma:
These generations don't have as many Jokers, so I've bundled them together.
This does come with some bugfixes:
- Vanillite line is now Blueprint Compatible
- Shiny Edition's +1 Booster Pack slot is now properly removed when destroying cards while skipping the dissolve/shatter animations, most notably present on non-Pokermon food jokers getting eaten


InertSteak:
Wow excellent! Thanks so much for the incredible job on this

PR #671, opened 2026-02-12
Refactor Scaling Jokers (Gen 9)

emmadenslemma:
Last refactor (Until I get to Bitter Rival and Ruins of Alph, both of which won't happen anytime soon), only includes Fidough and Charcadet lines

As for the changes that went into most of these refactors, here are a few notes to consider when making new Jokers (or refactoring the old ones)

- Returning `mult`, `chips`, or `Xmult` instead of `mult_mod`, `chip_mod`, or `Xmult_mod` handles all of the displaying automatically. It displays the amount, shows the correct colour, chooses the correct card to display it on, and more importantly automatically skips mult/chip values of 0 and Xmult values of 1 (Xmult below 1 still applies as usual)
- `context.after|before|joker_main` are all only called once, so we don't need to check for the cardarea being G.jokers. Additionally these are only called in the event of a hand going through (they are skipped for things like The Psychic when you play fewer than 5 cards) so `context.scoring_hand` is also guaranteed to be set.

There are a few more specific ones, like card areas will always have a `cards` field, and both cards and card areas will always have a `config` field, as they can't be instantiated without them, but those are less important to remember.

Concerning the `SMODS.scale_card` API for future use:
```lua
card.ability.extra.mult = card.ability.extra.mult + card.ability.extra.mult_mod
card_eval_status_text(card, 'extra', nil, nil, nil, {message = localize("k_upgrade_ex")})
```
is equivalent to doing
```lua
SMODS.scale_card(card, {
  ref_value = 'mult',
  scalar_value = 'mult_mod'
})
```

`SMODS.scale_card` will automatically display an "Upgrade!" popup with the default orange background.
To change the colour you can do 
```lua
SMODS.scale_card(card, {
  ref_value = 'mult',
  scalar_value = 'mult_mod'
  message_colour = G.C.MULT
})
```
For most cases this will be enough, but a few more options include
- `message_key = 'a_xmult'` to change "Upgrade!" to "X0.1" (uses the value of the modifier)
- `no_message = true` to disable it completely, for scaling a card silently (or for when you want to display a message elsewhere like with Scyther)
- `scaling_message`, to override the message completely

the last one is used like this: 
```lua
SMODS.scale_card(card, {
  ref_value = 'Xmult',
  scalar_value = 'Xmult_mod',
  scaling_message = {
    message = localize('poke_geomancy_ex'),
    colour = G.ARCS.LOC_COLOURS.fairy,
    sound = 'tarot1'
  }
})
```

Lastly, for Jokers with slightly more complex scaling, we have the `operation` parameter:
```lua
-- for decreasing a value:
SMODS.scale_card(card, {
  ref_value = 'mult',
  scalar_value = 'mult_minus',
  operation = '-',
  message_key = 'a_mult_minus'
})
-- for multiplying a value:
SMODS.scale_card(card, {
  ref_value = 'Xmult',
  scalar_value = 'Xmult_multi',
  operation = 'X'
})
-- or for some custom scaling, like Dachsbun:
SMODS.scale_card(card, {
  ref_value = 'chips',
  scalar_value = 'chip_mod',
  operation = function(ref_table, ref_value, initial, modifier)
    ref_table[ref_value] = initial + modifier + #find_pokemon_type("Fire") * 2
  end
})
```
That last one is necessary for anything more than a quick A = A + B, so you're going to run into it again

Hopefully this all made sense, I'll be back next week to change all `remove()` calls to use `SMODS.destroy_cards` instead

### DIFF
--- a/pokemon/pokejokers_32.lua
+++ b/pokemon/pokejokers_32.lua
@@ -27,25 +27,24 @@ local charcadet={
   blueprint_compat = true,
   eternal_compat = true,
   calculate = function(self, card, context)
-    if context.cardarea == G.jokers and context.scoring_hand then
-      if context.joker_main then
-        return {
-          message = localize{type = 'variable', key = 'a_mult', vars = {card.ability.extra.mult}}, 
-          colour = G.C.MULT,
-          mult_mod = card.ability.extra.mult
-        }
-      end
-      if context.after and not context.blueprint then
-        card.ability.extra.mult = card.ability.extra.mult + card.ability.extra.mult_mod
-        return {
-          message = localize('k_upgrade_ex'),
-          colour = G.C.MULT
-        }
-      end
+    if context.joker_main then
+      return {
+        mult = card.ability.extra.mult
+      }
+    end
+    if context.after and not context.blueprint then
+      SMODS.scale_card(card, {
+        ref_value = 'mult',
+        scalar_value = 'mult_mod',
+        message_colour = G.C.MULT
+      })
     end
     if not context.repetition and not context.individual and context.end_of_round and not context.blueprint then
       card.ability.extra.mult = card.ability.extra.mult_original
-      card_eval_status_text(card, 'extra', nil, nil, nil, {message = localize('k_reset'), colour = G.C.RED})
+      return {
+        message = localize('k_reset'),
+        colour = G.C.RED
+      }
     end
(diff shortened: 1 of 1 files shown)


InertSteak:
I'm extremely grateful for your work and the information provided. Thanks!

PR #677, opened 2026-02-21
Refactor Card Destruction

emmadenslemma:
Refactors the internals of `poke_remove_card` to use `SMODS.destroy_cards`, and fully deprecates the `remove` function alongside replacing all occurrences.


InertSteak:
Excellent work as always!

PR #679, opened 2026-02-21
Updated spanish localization

heyctf:
I updated my localization after a long time, and also came back to fix a lot of issues in the last version of it.


InertSteak:
Nice, thanks!

PR #680, opened 2026-02-22
add `pokecompat.lua` to handle deprecations

emmadenslemma:


emmadenslemma:
unsure about whether or not pokeconstants is necessary, some of these could be in the main pokermon.lua file, I'll look into that after adding a pokestages file to handle stage pools

InertSteak:
Yeah, that sounds good to me, I think it could go either way. Thanks as always!

PR #681, opened 2026-02-23
Custom Stage Pools

emmadenslemma:
Admittedly this is still kind of a nebulous change, as the functionality of `get_random_poke_key` can't quite be recreated with out of the box SMODS ObjectTypes.
Namely we're missing:
- Combining Stages/Rarities (for Mystery Egg)
- Filtering based on Types (for Corsola)
- Excluding specific Jokers (for Jirachi)

Due to the above this doesn't deprecate `get_random_poke_key` or `create_random_poke_joker`, but it still has some changes that I'd say are worthwhile in spite of the shortcomings:
- you can now specify "Pokemon" instead of "Joker" as a set when calling `SMODS.create_card`, to generate only Pokermon Jokers
- in the same vein all Stages have received their own pool, accessed by specifying "Stage " followed by the name of your Stage instead of "Joker"
- all Pokeball style items have been changed to use the new stage pools, which fixes an issue of CardSleeves changing the outcomes of RNG seeds when installed
- all Safari 'aux_poke' Jokers have received in_pool functions to remove them from the Safari pool
- all Safari Joker generators now use `SMODS.create_card({ set = 'Joker', rarity = 'poke_safari' })`

I should also mention that this approach does carry 2 notable bugs at the moment:
- Specifying a Rarity when using the Stage pools is ignored, due to how SMODS ObjectTypes handles rarity.
- Specifying "Common", "Uncommon", or "Rare" instead of 1, 2, or 3 when using the "Pokemon" pool crashes the game.

There's still a good chance that the correct way to do this is going back to `get_random_poke_key`. As always let me know if you have any changes that need to be made to this, or if you have any ideas on how to improve this (Especially to better facilitate Fields in the future)

### DIFF
--- a/functions/pokestages.lua
+++ b/functions/pokestages.lua
@@ -68,6 +68,9 @@ POKE_CUSTOM_POOLS = {
     return (stage == "Baby" or stage == "Basic")
         and (rarity == "Common" or rarity == "Uncommon" or rarity == "Rare")
   end,
+  ["Corsola"] = function(center)
+    return center.stage == "Basic" and center.ptype == "Water"
+  end,
 }
 
 poke_setup_custom_pool = function(pool)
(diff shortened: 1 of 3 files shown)


emmadenslemma:
Yeah actually hold off on merging this for now, I think I have a better way to do this

InertSteak:
Got it, no problem

PR #685, opened 2026-02-26
Updated Munna and Lampent line localization

heyctf:
Updated Munna, Musharna, Lampent and Chandelure description. Also Illumise's description slightly adjusted


InertSteak:
Thanks!

PR #687, opened 2026-03-01
Joker Generation fixes

emmadenslemma:
Some small changes that are important enough to merge without a big overhaul

- All Safari `aux_poke` Jokers have gotten `in_pool` methods to remove them from the Safari pool, similarly to how Jirachi Forms and Shaymin-Sky operates
- Safari Tag and Ruins of Alph's Safari reward now use the regular `SMODS.add_card` functionality
- `get_random_poke_key` has been changed from looking through `G.P_CENTERS` to use `G.P_CENTER_POOLS.Joker` instead, to fix a long running bug where installing CardSleeves would change the outcomes of Pokéball items


InertSteak:
That all sounds good, thanks!

PR #690, opened 2026-03-03
Added Mystery Dungeon daily challenge

SonfiveTV:
Mystery Dungeon challenge that is a fixed seed every day, letting all users play and compete! 

It currently gives a random Eternal Joker, and applies a random Pokermon deck effect too

Understand if this doesn't get accepted, it wasn't requested after all. I just thought it was cool and since I coded it I figured I'd give it a shot anyway


InertSteak:
Looking at this now, seems awesome!

InertSteak:
Ok, I changed a few things
* Fixed an issue where the deck was random everytime. Apparently apply happens before start_run. I simply moved everything inside the event call. 
* Excluded legacy decks from the pool of random decks. Maybe pending feedback we reenable them.  
* Changed to 12 hour time in the description. I figure more people are familiar with 12 hour. I can change it back again depending on feedback.

PR #692, opened 2026-03-04
Show Mystery Dungeon Challenge's randomly picked deck

emmadenslemma:
Admittedly the simplest way to do this would be replacing the sprite objects set in `Card:set_sprites`, but this at least comes with a nice "pop" effect visually.

Also allows you to do shaders to draw things like half challenge deck half custom sprite, the sky is the limit.

### DIFF
--- a/functions/pokedraw.lua
+++ b/functions/pokedraw.lua
@@ -100,12 +100,19 @@ SMODS.DrawStep {
   conditions = { vortex = false, facing = 'front' },
 }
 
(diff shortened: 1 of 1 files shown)


emmadenslemma:
Oops this is currently missing a check for whether or not the back sprite is actually the challenge deck's back sprite

InertSteak:
Ah I see, I'll merge it when you give the go ahead then. Looks good so far

PR #693, opened 2026-03-04
Updated spanish localization (Deoxys, Sharpedo line and challenge)

heyctf:
Updated version of my localization, with the Deoxys update, Sharpedo line, the new challenge and lots of fixes


InertSteak:
Nice, thanks!

PR #696, opened 2026-03-10
update mystery dungeon back sprite code to fix tilted cards

emmadenslemma:
*better late than never*

Changes the Mystery Dungeon draw code to replace the back sprites instead of drawing over them.

Still has the delay and visual pop when starting the challenge.


emmadenslemma:
You could make a case for having this code in the challenges file instead, but for now I've left it in `pokedraw.lua`

InertSteak:
Alright looks good to me. Thanks!

PR #699, opened 2026-03-12
fix reusables from packs

emmadenslemma:
Fixes a bug where items like Leftovers wouldn't get removed from the pool if used directly from packs


emmadenslemma:
Turns out, this is not so easily solvable. Probably have to get SMODS in on it.

PR #703, opened 2026-03-19
fix teeter dance localization key

emmadenslemma:
quick fix for spinda's popup message not working properly


InertSteak:
Looks good, thanks!

PR #719, opened 2026-04-12
Refactor Sun Stone

emmadenslemma:
Makes Sun Stone and other rank randomizer effects respect in_pool functions for custom ranks.

Currently clashes with Maelmc's Ogerpon implementation, where he uses a custom rank to let Stone Cards have ranks, this should free up other addons to add similar Jokers without having to patch out the function themselves


InertSteak:
Oh very nice work. Thanks!

PR #723, opened 2026-04-14
Add Attributes

emmadenslemma:
Adds support for the new `SMODS.Attribute` feature to every currently released Joker.

I've no doubt missed a couple of attributes, and we're sure to get more in the future, but this should at least function as a starting point for anyone wanting to use this feature

Shouldn't cause any issues on prior versions of SMODS but I have not tested it myself


InertSteak:
Oh wow, this is crazy impressive

PR #726, opened 2026-04-16
Misc. Refactors

emmadenslemma:
Refactors of
- Litwick, Lampent, and Chandelure
- Rival, Bitter Rival, and Champion
- Ruins of Alph
- Deino, Zweilous and Hydreigon

Litwick, Lampent, Rival, Bitter Rival, and Ruins of Alph (as well as the Drain mechanic in general) now use `SMODS.scale_card`
Deino and Zweilous have been cleaned up, while Hydreigon has had a partial revert to use the old card destruction animation.
Champion no longer says "Smell ya Later" and the Rival line in general no longer give bonus chance at money.

### DIFF
--- a/functions/energyfunctions.lua
+++ b/functions/energyfunctions.lua
@@ -290,16 +290,6 @@ ease_poke_dollars = function(card, seed, amt, calc_only)
         earned = earned + 1
       end
     end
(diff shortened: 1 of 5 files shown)


InertSteak:
Nicely done as always

PR #728, opened 2026-04-19
Fix Ruins Refactor

emmadenslemma:
Fixes a small goof I made while refactoring Ruins of Alph where it would create Unowns *before* The Mirror selected a target to turn into Ditto, always giving you a negative Ditto


InertSteak:
Ah I see, good catch!

PR #729, opened 2026-04-21
JokerDisplay - Skarmory bug fix

Definite-Internet-User:
Skarmory's joker display was bugged when used with Gigalith and I assume other jokers that retrigger held in hand cards.


InertSteak:
Alright good catch!

PR #730, opened 2026-04-23
re-revert hydreigon

emmadenslemma:
changes Hydreigon back to the less saucy version after reports of Hydreigon making gaps in your hand


InertSteak:
Right, I noticed reports of that. Thanks for the fix!

PR #731, opened 2026-04-24
Volbeat Perishable from true to false

Lemmanade0:
Swapped Volbeat to not be perishable as it is a scaling joker


InertSteak:
Good catch, thanks!

PR #734, opened 2026-04-26
Refactor Smeargle

emmadenslemma:
Yet another Smeargle Rewrite has hit the Pokermon mod

This one fixes a crash with Cryptid installed, and updates Smeargle to use `SMODS.blueprint_effect`.

### DIFF
--- a/functions/pokefunctions.lua
+++ b/functions/pokefunctions.lua
@@ -564,8 +564,9 @@ get_family_keys = function(card)
     end
   end
   if center.name == "smeargle" then
-    if card.ability.extra.copy_joker then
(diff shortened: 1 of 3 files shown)


InertSteak:
Looks good overall, but there is one oddity I noticed. 

When you copy a joker and then sell it, Smeargle retains the ability. However, if you then quit out of the run and relaunch it, Smeargle won't have that ability anymore. It should probably either loss the ability when the joker is sold or keep it upon quitting.

emmadenslemma:
Good catch! I forgot to check for whether the copied joker had been removed, it should be fixed now

PR #736, opened 2026-04-27
Updated es_419 and es_ES

heyctf:
Hello! Im updating again my spanish localization for the mod and latest updates, this includes everything after the Ralts and Meditite line update. Also, recently I had the idea to re-use my localization for the es_ES file since it hasnt had an update for a year now.

(Si algún español ve esto y quiere arreglar o aportar en la localización, adelante porque sé que habrán algunas cosas que me faltan corregir :])


InertSteak:
Nice, looks good. Thanks!

PR #738, opened 2026-04-29
Fix formatting in installation instructions

HolloFox:
Format installation instructions to denote start and end of path due to "and not" plus "or" could be misinterpreted as part of path string.


InertSteak:
Thanks!

PR #739, opened 2026-04-30
Refactor Star and Iridescent Hacker

emmadenslemma:
Refactors The Star and Iridescent Hacker so your debuff state updates if you Tera into/out of debuffed types in-between hands


InertSteak:
Excellent work!

PR #740, opened 2026-04-30
Refactor Duskull Line

emmadenslemma:
Refactor Duskull line to fix the "Phantom Card" bug happening with Dusclops


InertSteak:
One day surely, steamodded will fix this

PR #741, opened 2026-04-30
fix type stickers in shop/collection crashing the game

emmadenslemma:
Fixes a bug introduced with the Star rework, that crashes by trying to debuff cards like Weird Tree in the shop/collection


InertSteak:
Ah good catch, thanks!

PR #743, opened 2026-05-03
Refactor Simisear and Anorith line

emmadenslemma:
Refactors Simisear to use `context.destroy_card` to fix yet another "Phantom Card" bug
Also includes a free refactor of Anorith and Armaldo

The last remaining Jokers to use `poke_remove_card` are:
- Slugma and Magcargo
- Anorith and Armaldo
- Huntail

None of which are giving me Phantom Card bugs in testing, possibly because they don't destroy played cards


InertSteak:
Alright, makes sense. Thanks!

PR #744, opened 2026-05-04
Split Rival into 3 separate Jokers

emmadenslemma:
Splits the Rival line into 3 separate Jokers to let separate things like sticker compatibility and attributes

Rival and Bitter Rival have had `auto_sticker` added for the sticker previous evo functionality, and Champion is now eternal compatible

Bitter Rival and Champion still do not show up in the collection outside of Pokédex View


InertSteak:
Great work as always

PR #747, opened 2026-05-06
Switch to using `mod.menu_cards()`

emmadenslemma:
Uses the new built-in SMODS functionality to make menu cards instead of patches.
Additionally chooses the Unown Title Card sprite that matches your current Unown sprite in use, and includes a 1/4096 change for the card to be shiny

Not included here but planned are delaying the shiny application (with added visual flair and sound effects) and replacing Multiplayer's custom card with Unown 'E'


InertSteak:
Nice, the lest patches we have to deal with, the better

PR #748, opened 2026-05-07
Misc fixes

emmadenslemma:
Includes a fix for the menu card not having its soul sprite rendered on new saves, and makes cards in the Individual Sprites menu no longer need to be discovered to see their sprites (The description will still say "Not Discovered" but the sprite will be rendered)


InertSteak:
Nice, thanks!

PR #749, opened 2026-05-07
Remove references to G.ASSET_ATLAS

emmadenslemma:
Changes all references to `G.ASSET_ATLAS` to instead use `SMODS.get_atlas`

Admittedly kind of a find and replace job, since SMODS' function just checks the animated atlas in case it can't find it in the asset atlas, it should be safe to replace

### DIFF
--- a/editions/editions1.lua
+++ b/editions/editions1.lua
@@ -41,7 +41,7 @@ local shiny = ({
       end
     end,
     on_load = function(card)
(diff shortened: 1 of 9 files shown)


emmadenslemma:
Should probably also mention this is a fix for an Electrode-related crash from switching it to using ANIMATION_ATLAS

InertSteak:
Alright, sounds good to me

PR #751, opened 2026-05-12
add gen 4 origin forms to dex_order

emmadenslemma:
self serving fan additions PR, just adds the origin forms to the dex order list


InertSteak:
That's a good thing in my book!

PR #753, opened 2026-05-13
remove mutation from poke_create_art_display_card

emmadenslemma:
very minor coding change to the inner workings of the art credits menu, almost certainly won't ever actually matter but it will keep me up at night if I leave it unchanged


InertSteak:
I feel you there

PR #754, opened 2026-05-15
Fix multiple Machamps taking away all your hands

emmadenslemma:
- Moves `poke_get_consumables` into pokeutils.lua, and `poke_drain_chips` into pokefunctions.lua
- Adds a new helper function, `poke_ease_hands_played`. This function works identically to `ease_hands_played` except when using a negative modifier, where it it won't let your hands drop below 1
- Adds a new buffer variable `G.poke_hands_buffer`. This is automatically set and cleared whenever `ease_hands_played` is called without specifying instant, and is used in the new `poke_ease_hands_played` function
- Updates all Water Starters, the Machop line, and Mega Ampharos to use the new `poke_ease_hands_played` when removing hands
- Removes the exception for The Needle in Mega Ampharos' code, since `poke_ease_hands_played` runs after the boss blind it will take the newly set buffer variable into account

### DIFF
--- a/functions/pokefunctions.lua
+++ b/functions/pokefunctions.lua
@@ -1416,18 +1416,19 @@ poke_can_save_consumable = function(card)
       or (card.config.center.saveable)
 end
 
-poke_get_consumeables = function(set)
-  local consumeables = {}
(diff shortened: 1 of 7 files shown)


InertSteak:
Great, thanks!

PR #763, opened 2026-05-15
Bunnelby

ThorsGirdle:
For real this time. 

I think they're slightly too strong but setting the probability to 1 in 3 felt too bad to me. I'm also not 100% on Bunnelby's evolution text, but I couldn't think of a better way to phrase it. Other than that I'm very fond of how this design encourages digging through your deck.

Also I think they only trigger once if two last Ranks are drawn at once, so it might need some refactoring if that's something you want it to do.


InertSteak:
Congrats, this has been accepted!

PR #764, opened 2026-05-16
Add Klink line

ThorsGirdle:
Honestly these guys might be the best balanced Jokers I've made so I don't really recommend buffing them. I'm particularly fond of how Klinklang becomes an alternative Acrobat, but does it in a may that's heavy on interactions and synergies.

I don't think I ever got around to making them properly blueprint combatable though.


InertSteak:
This has been accepted!

PR #767, opened 2026-05-16
refactor tyrogue

emmadenslemma:
While I haven't been able to confirm it, I am wagering that storing playing card references in `card.ability.extra` is our culprit in the Multiplayer Crash Saga, so hopefully this is the last we see of it


InertSteak:
Oh yeah that's definitely it. Tyrogue was coded a while ago and was quite...sloppy. Thanks for the refactor!

PR #768, opened 2026-05-16
Adding Elite Stake

Eternalnacho:
The Elite Stake from my add-on! Difficulty-wise it ends up being roughly analogous to Gold Stake in vanilla balatro thanks to the relative power level of Pokermon's jokers.

It adds a +2 ante win requirement, meaning a winning run ends at ante 10, and it changes the last *4* antes to be finisher boss blinds (Verdant Leaf, Iridescent Hacker, etc.) rather than just the last ante (hence the name).

As an aside for Atlas Boy™, please do whatever you want with the atlas files, i just left them separate for simplicity's sake


InertSteak:
I think you might have forgotten a block of code, G.GAME.modifiers.elite4 is set to true but never referenced by anything else I don't think.

Eternalnacho:
apparently i did. which is hilarious because it's the part that, y'know, makes the stake work.

InertSteak:
Yeah I was messing around with it, got to Ante 10 and then said "wait a second" lol

InertSteak:
Congratulations, this PR has been accepted!

PR #784, opened 2026-05-19
Moving values to keep in evolution to their respective jokers

Eternalnacho:
Teensiest bit of spring cleaning: the variables moved to poke_custom_values_to_keep are only being used by those specific jokers, and it feels odd to have them be generic exceptions in backend_evolve in light of that.

Also gourgeist's check for poke_can_set_sprite isn't using a card currently despite the function requiring one


InertSteak:
Yeah that all sounds good to me

PR #785, opened 2026-05-19
null-checking challenge.rules.custom

Eternalnacho:
without checking if rules.custom is an empty table, every challenge that didn't have any custom rules would crash (vanilla challenges included)


InertSteak:
Yeah good catch

PR #786, opened 2026-05-19
Smeargle tooltip fix

Maelmc:
- Fixed bugged display when the key doesn't exist by checking if loc_vars returns a key (e.i. gourgeist)
- Future fix for the next SMODS release, which allows to properly display variables that are in the joker's name

Tested in both 1620a and the newest dev version and both work


InertSteak:
Very nice work

PR #787, opened 2026-05-20
Refactor Zorua and Zoroark

emmadenslemma:
Zorua and Zoroark Refactor!

This refactor includes:
- Dark Tera while in the shop!
- No longer disguises as Dark types!
- Has "Compatible/Incompatible" in the description!
- Copies Animated Sprites correctly!

Also includes changes to `poke_is_in_collection` so it might be worth checking if that's still working properly everywhere.
It still doesn't work with Unown Swarm.


InertSteak:
That was fun to test indeed. I wasn't able to break it thankfully. Great work!

PR #790, opened 2026-05-21
Rework Tera Orb

emmadenslemma:


InertSteak:
Looks like it works as intended

PR #793, opened 2026-05-22
fix poke_copy_sprite crashing

emmadenslemma:
fixes a zoroark-related crash posted in fan-additions


InertSteak:
Thanks as always!

PR #794, opened 2026-05-22
perishable_compat to false for Swablu, Altaria, Baltoy, Claydol, and Diggersby

Definite-Internet-User:
They're non-retroactive scaling jokers so they probably shouldn't be able to have perishable.


InertSteak:
Whoops, forgot to merge this. Thanks!

PR #795, opened 2026-05-23
fix pumpkaboo sprite pos

emmadenslemma:
fix pumpkaboo sprite indexes not being updated after the vivillion addition


InertSteak:
Awesome, thanks!

PR #801, opened 2026-05-28
Added "enhance_cards() function"

Numbuh214:
Been awhile, huh...?

I've just been playing modded Balatro again, and I thought I'd add a function to make it easier to create visuals more in line with enhancing cards in vanilla Balatro. I've showed it off via recoding Arcanine, who was my example in all this.

### DIFF
--- a/functions/uifunctions.lua
+++ b/functions/uifunctions.lua
@@ -111,7 +111,7 @@ poke_blueprint_compat_ui = function(copy)
   }
 end
 
-enhance_cards = function(cards, enhancement, joker, message, _delay, sound)
+pokermon.enhance_cards = function(cards, enhancement, joker, message, _delay, sound)
   local i = 1
   if type(cards) == 'nil' then
     sendDebugMessage("Attempted to enhance 0 cards.")
--- a/functions/uifunctions.lua
+++ b/functions/uifunctions.lua
@@ -111,15 +111,15 @@ poke_blueprint_compat_ui = function(copy)
   }
 end
 
-pokermon.enhance_cards = function(cards, enhancement, source, message, _delay, sound)
+pokermon.enhance_cards = function(cards, enhancement, source, _message, _delay, sound)
   if type(cards) == 'nil' then
     sendDebugMessage("Attempted to enhance 0 cards. Function failed.")
     return
   end
   if type(cards.is) == 'function' and cards:is(Card) then
     sendDebugMessage("Single card passed. Converting to table and continuing.")
-    local ctable = {cards}
-    cards = ctable
+    pokermon.enhance_cards({cards}, enhancement, source, _message, _delay, sound)
+    return
   end
   
   local default = type(_delay) ~= 'number'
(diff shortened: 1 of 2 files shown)


functions/uifunctions.lua
```diff
@@ -111,6 +111,42 @@ poke_blueprint_compat_ui = function(copy)
   }
 end
 
+enhance_cards = function(cards, enhancement, joker, message, _delay, sound)
```
emmadenslemma: `enhance_cards` should be `pokermon.enhance_cards` in accordance with the upcoming prefix refactor

Numbuh214: I believe that fixes the issue as requested. If it does not, please let me know how else I could fix it.

emmadenslemma:
Optional changes to implement (Partly up to taste, probably won't be enforced):
- `type(cards) == 'nil'` should also check for empty tables (To differentiate empty tables from Card objects, use `type(cards.is) == 'function' and cards:is(Card)`)
- The Debug messages on successful interactions should be removed (Converting single card to table, iterating through cards etc)
- Should add a Debug message on enhancement key not being found in `G.P_CENTERS`
- `joker` parameter could be renamed for clarity, consider `trigger_obj`, `caller`, `source` or similar.

Numbuh214:
That should do it. I check for nil objects first specifically because I make sure to catch whether or not the user passed a single card without wrapping it in a table, because that's a common mistake in Joker coding that I wanted to account for. So essentially:

1. Is the "cards" table nil? If so, exit the function.
2. Is the "cards" table actually a single card?
  - If so, wrap it in a table and continue the function.
  - If not, exit the function.

That all said, I believe I've implemented all the other suggestions. I commented out my original debug messages, so they can remain for the sake of clarity when people inspect the code.

emmadenslemma:
I'll give it another look over. I do think the debug messages should stay when sending nil/an empty table through, to let developers know more quickly when they're using the function incorrectly

functions/uifunctions.lua
```diff
@@ -111,6 +111,48 @@ poke_blueprint_compat_ui = function(copy)
   }
 end
 
+pokermon.enhance_cards = function(cards, enhancement, source, message, _delay, sound)
+  local i = 1
+  if type(cards) == 'nil' then
+    --sendDebugMessage("Attempted to enhance 0 cards.")
+    return
+  end
+  if type(cards.is) == 'function' then
```
emmadenslemma: `cards.is` needs to be called with `Card` after checking that it exists.
the full statement should look something like
```lua
if type(cards.is) == 'function' and cards:is(Card) then
```

functions/uifunctions.lua
```diff
@@ -111,6 +111,48 @@ poke_blueprint_compat_ui = function(copy)
   }
 end
 
+pokermon.enhance_cards = function(cards, enhancement, source, message, _delay, sound)
+  local i = 1
```
emmadenslemma: `i` is unused

functions/uifunctions.lua
```diff
@@ -111,6 +111,48 @@ poke_blueprint_compat_ui = function(copy)
...
+    v:set_ability(G.P_CENTERS[enhancement], nil, true)
+  end
+  G.E_MANAGER:add_event(Event({
+    trigger = 'before',
+    delay = default and 0.7 or _delay,
+    func = function()
+      for k, v in pairs(cards) do
+        --sendDebugMessage("Card "..i.." enhancement: "..tostring(v.config.center_key))
+        v:juice_up()
+        i = i + 1
```
emmadenslemma: `i` is unused

functions/uifunctions.lua
```diff
@@ -111,6 +111,48 @@ poke_blueprint_compat_ui = function(copy)
...
+  if type(cards) == 'nil' then
+    --sendDebugMessage("Attempted to enhance 0 cards.")
+    return
+  end
+  if type(cards.is) == 'function' then
+    --sendDebugMessage("Single card passed. Converting to table.")
+    local ctable = {cards}
+    cards = ctable
+  end
+  elseif type(cards[1].is) ~= 'function' then
```
emmadenslemma: `type(cards[1].is) ~= 'function'` needs to be `type(cards[1].is) ~= 'function' or not cards[1]:is(Card)`

emmadenslemma (changes requested):

emmadenslemma:
This looks to be in order 👍

InertSteak:
Looks good to me

PR #808, opened 2026-05-30
Updated spanish localization (up to second flareon update lolol)

heyctf:
Quick update for my localization, might fix some issues later :p


InertSteak:
Looks good, thanks

PR #809, opened 2026-05-31
fix tangela and tangrowth scoring messages firing twice

emmadenslemma:


InertSteak:
Ah good catch, thanks!

PR #813, opened 2026-06-03
Fix Tera Orb + Copy interaction where sprite doesn't update correctly

Eternalnacho:
Per Numbuh214's observation on discord, should hopefully fix the interaction between tera orb and copy effects (ho-oh, etc.) where the sprite doesn't match the tera type


InertSteak:
Good catch, thanks!

PR #818, opened 2026-06-12
fix display cards opening pokedex view

emmadenslemma:
fixes a small oversight on my part, oopsie


InertSteak:
No worries, its all good

PR #820, opened 2026-06-13
Fix Rellor line crash with JokerDisplay

Definite-Internet-User:
Fix a crash if you pick up a Rellor/Rabsca while having used zero items when using Joker Display. Also it now scales with energy which I somehow forgot to do initially.


InertSteak:
Good catch

PR #821, opened 2026-06-15
Missed a spot in pokecompat

Eternalnacho:
oopsie

(moves the load call in the main file for pokecompat.lua down so it catches the spriteload functions, and also fixes the copy + paste error)


InertSteak:
Looks good!

PR #824, opened 2026-06-19
add a create_held_item function (and use it for the jokers that hold...

Eternalnacho:
…things)


InertSteak:
Excellent, I appreciate it!

PR #825, opened 2026-06-20
so i forgot to null check previous evo overrides...

Eternalnacho:
beautifly and jokers with specific overrides for the get_X_evo functions would crash when stickering because the recursive function bit assumed the *lower* evos *also* had previous evo overrides and they don't. oops.


InertSteak:
Oh nice, I was going to look into that, but you beat me to it!

PR #828, opened 2026-06-23
JokerDisplay - Relicanth fix money bug

Definite-Internet-User:
Money was only showing when there was 3+ fours in poker hand but not 2+ fours


InertSteak:
Oh yeah good catch, I forgot about that haha

PR #831, opened 2026-06-25
Prefix Colours

emmadenslemma:
Prefixes Pokermon-specific colours, to ensure we don't run into any collisions. Additionally removes superfluous instances of Item and Safari colours.

Instead of using `G.ARGS.LOC_COLOURS` in the future:
- `G.ARGS.LOC_COLOURS.item` -> `G.C.SECONDARY_SET.poke_item`
- `G.ARGS.LOC_COLOURS.safari` -> `G.C.RARITY.poke_safari`
- `G.ARGS.LOC_COLOURS[ptype]` -> `pokermon.colours[ptype]`

`pokermon.colours` contains all unprefixed colours (except `item` and `safari`)

For colours in localization files, all colours are now prefixed with `poke_` (ex. `{C:pink}` -> `{C:poke_pink}`)

(localization files were edited by mass find and replace using regex)


InertSteak:
Excellent, now we're feeling like a real mod lol

PR #833, opened 2026-06-25
refactor `type_tooltip` functionality and deprecate it

emmadenslemma:
Kills `type_tooltip` and instead patches in energy/mega tooltips.


InertSteak:
Excellent!

PR #834, opened 2026-06-26
Fix Jirachi Booster crashing

emmadenslemma:
Fixes the type_tooltip refactor not accounting for card being nil


InertSteak:
Ah so that's what it was

PR #835, opened 2026-06-27
missingno crash fix and code update

Eternalnacho:
Fixing the prefixing crash from missingno's item creation bits (missed that in the refactor whoops) and updating missingno's Tag code to better utilize SMODS.poll_object, which saves us the trouble of manually generating the tag and setting Orbital Tag specifically, etc.


InertSteak:
Excellent work

PR #838, opened 2026-07-01
Use poll_evo_item function for pickup keys

Eternalnacho:
Super niche PR, Linoone (and other pickup jokers) can crash if they attempt to generate an evo item from a fan addition without the "poke" prefix. Fortunately there's a handy poll_evo_item function already in use for pocket packs that doesn't have this issue, so i moved that function and applied it in both places.


InertSteak:
I love super niche PRs! Good work on this one

PR #845, opened 2026-07-09
Sprite Toggle crash fix

emmadenslemma:
also includes a fix for the gmax charizard and gmax blastoise sprites being swapped in pokespriteload


InertSteak:
Thanks!

PR #850, opened 2026-07-13
dragon scale bug? fix

Eternalnacho:
Was looking at item / energy related code and stumbled into dragon scales not displaying their "on use" card eval status text message properly and instead showing a blank square, so I figured I'd just tweak the code a little. Shouldn't change any functionality of anything, just using more recent code


InertSteak:
Thanks, looks good to me!

PR #851, opened 2026-07-13
Pumpkaboo line sticker scaling fix + create_consumeable event timing change

Eternalnacho:
Pumpkaboo sticker thing is self-explanatory, just slightly adjusting the way the scaling itself happens

create_consumeable is the exact same function as create_held_item, just with two new arguments for whether the card should materialize in an event or not and for which card should display the "+1 X" message

(also I'm sneaking in a thing for the energy tooltip and e_limit_up but that one isn't important shhhhh)


InertSteak:
Looks good!

Issue #852, opened 2026-07-14
Crash (Maybe Gourgeist?)


TCJLee:
Oops! The game crashed:
[SMODS Pokermon "pokemon/pokejokers_24.lua"]:174: attempt to index global 'v' (a nil value)

Additional Context:
Balatro Version: 1.0.1o-FULL
Modded Version: 1.0.0~BETA-1814a-STEAMODDED
Optional features enabled: quantum_enhancements
LÖVE Version: 11.5.0
Lovely Version: 0.9.0
Platform: Windows
Steamodded Mods:
    1: Trans Spectrals by RadicaAprils, AutumnMood (it/she/they) [ID: TSpectrals, Version: 1.2]
    2: Pokermon by InertSteak, Please See Credits Page [ID: Pokermon, Priority: -1, Version: 3.8.1-0713a, Uses Lovely]
    3: BalaGay by Nether [ID: balagay]
    4: Handy by SleepyG11 [ID: Handy, Version: 1.5.1p, Uses Lovely]
    5: JokerDisplay by nh6574 [ID: JokerDisplay, Priority: -1000000000, Version: 1.10.7, Uses Lovely]
Lovely Mods:
    1: BMM-Compat

Stack Traceback
===============
(3) Lua local 'func' at file 'pokemon/pokejokers_24.lua:174' (from mod with id Pokermon)
Local variables:
 (*temporary) = nil
 (*temporary) = nil
 (*temporary) = string: "attempt to index global 'v' (a nil value)"
(4) Lua field 'filter' at file 'functions/pokeutils.lua:203' (from mod with id Pokermon)
Local variables:
 list = table: 0x416a5850  {1:table: 0x428afee0, 2:table: 0x415e5510, 3:table: 0x41a8f858, 4:table: 0x411dcef0, 5:table: 0x41d7c748}
 func = Lua function '?' (defined at line 174 of chunk [SMODS Pokermon "pokemon/pokejokers_24.lua"])
 new_list = table: 0x4178e8f0  {}
 (for generator) = C function: next
 (for state) = table: 0x416a5850  {1:table: 0x428afee0, 2:table: 0x415e5510, 3:table: 0x41a8f858, 4:table: 0x411dcef0, 5:table: 0x41d7c748}
 (for control) = number: nan
 _ = number: 1
[...]

PR #853, opened 2026-07-14
gourgeist crash fix (whoops)

Eternalnacho:
missed a spot with the filter function, should fix the crash from issue #852


InertSteak:
No worries haha

PR #856, opened 2026-07-17
2 Teensy Optimizations for create_consumeable and Pumpkaboo line

Eternalnacho:
create_consumable fix is preventing the card spawning noise from stacking on itself (not super noticeable until you start having game states running at 4fps but was still something of note)

pumpkaboo and gourgeist in small form kinda freak out when you hover over them? and it's my fault because of how I did the sticker fix, so I made a little hook to fix the problem


InertSteak:
Alright, that looks good to me!

PR #857, opened 2026-07-18
Voucher ownership thingy for interest cap reasons

Eternalnacho:
Was asked to ship this one over so here it is


InertSteak:
Thank you very much!

PR #858, opened 2026-07-19
Showman compat fix

Maelmc:
Also moved the showman check in ditto's in pool func since I assume it was at the wrong place, since checking at the end is useless


InertSteak:
Thanks!

PR #859, opened 2026-07-19
Using SMODS.find_card instead of find_joker where applicable

Eternalnacho:
Some of this might conflict with mael's earlier PR since they were meant to address the same thing initially, but it also replaces instances of find_joker with SMODS.find_card


InertSteak:
Thanks!

PR #861, opened 2026-07-20
refactor xmult_mod returns

emmadenslemma:
Miscellaneous fixes pertaining to `Xmult_mod` returns, including Gholdengo no longer activating at 1.0 Xmult, and jokers like Charizard and Typhlosion now using the appropriate sound effects.

Also refactors every instance of Jokers returning `Xmult_mod` values.


InertSteak:
Thank you so much for this. That's something that always bugged me lol

PR #862, opened 2026-07-21
Patch display cards

emmadenslemma:
self serving patch to add card params to display cards (needed for fan addons,) but also adds glossary_ignore so it doesn't try to open the glossary while in the art credits menu


InertSteak:
Nice, looks good!

PR #863, opened 2026-07-21
depleted fix for ranks not in pool / non-default ranks

Eternalnacho:
This is mostly for fan-addon reasons - currently with the way deplete and pokermon.get_depleted work (which is really just relicanth atp), ranks that don't *start* in your deck count toward the depleted count because there are zero of them in deck by default, so this is just a bandage fix for that


InertSteak:
Hmm, that's a good idea, but could you possibly wrap the inner block of the for statement instead of using a goto?

Eternalnacho:
> Hmm, that's a good idea, but could you possibly wrap the inner block of the for statement instead of using a goto?

like this?

InertSteak:
Yup that works, thanks!

PR #864, opened 2026-07-22
let smeargle copy other smeargles

emmadenslemma:


emmadenslemma:
figured I should clarify this one: Vanilla Balatro already has a recursion check for Blueprint effects, this simply adds one for the Smeargle Tooltip (which was previously causing overflow errors)

InertSteak:
Oh nice, I didn't even realize that was an issue. Thanks!

PR #865, opened 2026-07-23
Pumpkaboo Discard Counter Fix

Eternalnacho:
Dunno how universal it is, pumpkaboo line's "jacks discarded" counter wasn't incrementing properly on my end and I saw a separate bug report about it, so here's a fix just in case


InertSteak:
Alright, looks good to me!

PR #867, opened 2026-07-24
update create_consumeable to mimic vanilla behavior with banned keys

Eternalnacho:
should be what it says on the tin, just getting the set beforehand so that it creates a random consumable of the same set in case the key is banned


InertSteak:
Nice, thanks!

PR #868, opened 2026-07-25
create_consumable crash fix

Eternalnacho:
missed a thing with args.set, if it was set beforehand we make sure it's not overwritten when checking for it later


InertSteak:
Ah I see, I missed that. Thanks!

PR #869, opened 2026-07-26
ninetales held item should be negative

Eternalnacho:
whoopsie, i think that one's on me


InertSteak:
It's all good, happens to the best of us

PR #870, opened 2026-07-27
move pool toggles into `context.modify_weights`

emmadenslemma:
Using the new SMODS feature of object weights, this simply moves our pool functions out of patches and into context calls.

Ideally shouldn't change any regular functionality, but does give us 2 small benefits:
- pool toggles will now also apply to joker generation using attributes
- we can now use `weight` to make certain consumables/jokers/enhancements/etc more or less likely to appear (base weight is 10)


InertSteak:
Awesome, looks good to me. The less patches we have to rely on the better.

PR #872, opened 2026-07-30
Split getting + applying values to keep / custom values to keep into a separate function

Eternalnacho:
This is kind of a selfish PR, and functionally nothing should change, but I wanted specifically this functionality for a fan addition joker (passimian lol) so uh... please? let me know if things don't work as intended

the only other thing i changed is taking out some unnecessary state-setting, since we set the new children.center role to glued anyway


InertSteak:
Sure, looks good to me.

PR #875, opened 2026-08-01
Refactor Gastly line

emmadenslemma:
Full rewrite of Gastly/Haunter and partial rewrite of Gengar, should hopefully fix the Gastly Kamikaze situation once and for all.
Also touches up the animations


InertSteak:
Looks good!

PR #880, opened 2026-08-11
Type tooltip removal

Maelmc:


InertSteak:
Thanks, this is a big help

PR #881, opened 2026-08-11
Fixed Axew line to work like all nature jokers

Maelmc:
All Nature jokers are in the form of {{nature 1}, {nature 2}} etc

Axew line was {nature 1, nature 2} which was causing issues for me specifically, but it's also just good in general to normalize it


InertSteak:
That's because ranks comprise two components in Balatro, but yeah standardization can't hurt

PR #883, opened 2026-08-12
Refactor Gimmighoul

emmadenslemma:
Refactors Gimmighoul and adds a parameter to `pokermon.add_shop_card` to make the resulting card free

This version retains the behaviour of Gimmighoul counting both earning and spending money. Let me know if we should take this opportunity to make it just earning instead

Given that [Talisman has been frozen](https://github.com/SpectralPack/Talisman) a little under a month ago, I've only tested this with its replacement Amulet


InertSteak:
Awesome, thanks!

PR #884, opened 2026-08-14
Attribute stages n types

emmadenslemma:
Bit of a hefty change, this Pull Request:
- Adds new (automatically defined) attributes for Stages and Types 
- Refactors everything except Mystery Egg, Zorua Disguise, and Randomizer mode to use attributes for Joker generation
- Changes the Hazards Allowed toggle to instead look for the 'hazards' attribute
- Adds Attributes to Baltoy and Claydol
- Cleans up pokeconstants.lua and pokermon.lua, moving Attribute declarations out of the latter and helper functions out of the former

Example of using this new way of generating specific stage/type jokers:
```lua
SMODS.add_card({
  set = 'Joker', -- important for challenge bans reasons
  attributes = {'stage_basic', 'water_type'}, -- does exactly as you'd expect.
  -- union = true, -- for if you want to get all basic pokes AND all water types, instead of only jokers that are both
  rarity = false, -- turns off rarity polling! this is important for smaller pools or things that generate safari jokers
  -- allow_legendaries = true, -- this *should* be on by default if using the 'stage_legendary' pool, but appears to be bugged 
  key_append = 'custom_rng_seed'
})
```

### DIFF
--- a/consumables/00_mart_balls.lua
+++ b/consumables/00_mart_balls.lua
@@ -18,7 +18,7 @@ local pokeball = {
     pokermon.set_spoon_item(card)
     G.E_MANAGER:add_event(Event({trigger = 'after', delay = 0.4, func = function()
         play_sound('timpani')
-        SMODS.add_card({ set = 'Joker', attributes = {'stage_basic'}, key_append = 'pokeball' })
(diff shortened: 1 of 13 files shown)


InertSteak:
Awesome, looks good!

PR #885, opened 2026-08-15
Smeargle tooltip adjustment to remove copy_table call

Eternalnacho:
So Ortalab has a patch like the one I'm replacing the loc_vars_replacement patch with - it lets you use card ui properly in info_queues. The upshot of this is that we now no longer need to call copy_table on the copy's ability table, meaning we can now use smeargle recursion in tooltips, along with all of the other places we couldn't before (theoretically).


InertSteak:
Sounds sweet, thanks!

PR #886, opened 2026-08-17
update mystery egg to use attributes

emmadenslemma:
yes, this is gobbledygook but it works.

I posted a bug report to the SMODS discord about attributes not working with rarities, if that gets fixed we can simplify this significantly, but until then this is what we're stuck with


InertSteak:
Whoops, I totally missed this pr somehow

PR #888, opened 2026-08-21
Added Energy & Sell value UI + config (2nd attempt)

SonfiveTV:
I had these in a small QOL mod and figured I would share here. Adds 2 toggles which are disabled by default
One to add the current Energy count underneath a Joker's description, the other to add it's sell value there Energy value being displayed underneath the Joker has been quite nice to play with, and the sell value is quite useful with Pokermon

image
image
image

### DIFF
--- a/localization/en-us.lua
+++ b/localization/en-us.lua
@@ -7322,18 +7322,18 @@ return {
                 "the collection",
               }
             },
-            poke_energy_count_tooltip = {
+            poke_energy_ui_tooltip = {
               name = "Energy Count",
               text = {
                 "Displays a Joker's {C:poke_pink}Energy{}",
-                "count in its description",
+                "count under its description",
               }
             },
-            poke_sell_value_tooltip = {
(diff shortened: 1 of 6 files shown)


emmadenslemma:
UI functions need to be local, or be part of the `pokermon.` namespace.

SonfiveTV:
I think that should be fine now

localization/en-us.lua
```diff
@@ -7322,6 +7322,20 @@ return {
                 "the collection",
               }
             },
+            poke_energy_count_tooltip = {
+              name = "Energy Count",
+              text = {
+                "Displays a Joker's {C:poke_pink}Energy{}",
+                "count in it's description",
```
emmadenslemma: "it's" should be changed to "its" for possessive (same for both tooltips)

SonfiveTV: This specifically should also be fine now

InertSteak:
Awesome, looks great now

PR #889, opened 2026-08-23
fix mystery egg crash

emmadenslemma:
fixes a mystery egg crash from calling `SMODS.poll_object` before the playing cards table is instantiated.

additionally stops RNG from being processed while viewing in the collection (TMJ will still trigger RNG calls)


emmadenslemma:
fix for https://discord.com/channels/1303403397457903658/1540827125970641086

InertSteak:
Thanks! I should have caught that but didn't

PR #892, opened 2026-08-27
add support for poke evo overrides with 1 final evo

emmadenslemma:
not technically a bugfix, but I did notice beautifly and dustox were both inside their own tables in `POKE_EVO_OVERRIDES`

this removes the need for that.


InertSteak:
Oh yeah I think that was a bandaid fix for some bug. Thanks!

PR #895, opened 2026-08-29
Fixed the undiscovered key of item and energy

Maelmc:


InertSteak:
Ah I was just about to look into that, nice. Thanks!

PR #897, opened 2026-09-01
fix sticker previous evo

emmadenslemma:
fix for the patch target changing with the newest release of SMODS


InertSteak:
Nice, thanks!

PR #898, opened 2026-09-06
Fix Unown Swarm squishing

emmadenslemma:
fix for https://discord.com/channels/1303403397457903658/1545831381282852864

comes with the side effect of Unown Swarm having a slightly larger hitbox than usual


InertSteak:
Thanks!

This will surely be the last fix needed for Unown Swarm

PR #899, opened 2026-09-06
Fix Unown Swarm being obtainable with Safari tags

emmadenslemma:
it wasn't the last fix :(


InertSteak:
Dang, we'll get them next time

